PR #14, opened 2015-09-30
support homebrew

jackpot51:


esell:
Hey, just an FYI this isn't quite working yet :( Everything builds but doesn't startup in virtualbox. Trying to track down what the issue is...

jackpot51:
What version of rustc are you using?

esell:
rustc version: `rustc 1.5.0-nightly (6108e8c3e 2015-09-28`

Seems like it's not happy with the linking portion, here is a sample:

		i386-elf-ld -m elf_i386 -o filesystem/apps/viewer/viewer.bin -T src/program.ld build/`basename apps/viewer/viewer`.rlib
		i386-elf-ld: warning: cannot find entry symbol _start; not setting start address


And looking at the output via `readelf` it doesn't look quite right:

		$ greadelf -a terminal.bin
		ELF Header:
  		Magic:   7f 45 4c 46 01 01 01 00 00 00 00 00 00 00 00 00
  		Class:                             ELF32
  		Data:                              2's complement, little endian
  		Version:                           1 (current)
  		OS/ABI:                            UNIX - System V
  		ABI Version:                       0
  		Type:                              EXEC (Executable file)
  		Machine:                           Intel 80386
  		Version:                           0x1
  		Entry point address:               0x0
  		Start of program headers:          0 (bytes into file)
  		Start of section headers:          80 (bytes into file)
  		Flags:                             0x0
  		Size of this header:               52 (bytes)
  		Size of program headers:           0 (bytes)
  		Number of program headers:         0
  		Size of section headers:           40 (bytes)
  		Number of section headers:         4
  		Section header string table index: 1

		Section Headers:
  		[Nr] Name              Type            Addr     Off    Size   ES Flg Lk Inf Al
  		[ 0]                   NULL            00000000 000000 000000 00      0   0  0
  		[ 1] .shstrtab         STRTAB          00000000 000034 00001b 00      0   0  1
  		[ 2] .symtab           SYMTAB          00000000 0000f0 000020 10      3   1  4
  		[ 3] .strtab           STRTAB          00000000 000110 000008 00      0   0  1
		Key to Flags:
  			W (write), A (alloc), X (execute), M (merge), S (strings)
  			I (info), L (link order), G (group), T (TLS), E (exclude), x (unknown)
  			O (extra OS processing required) o (OS specific), p (processor specific)

		There are no section groups in this file.

		There are no program headers in this file.

		There are no relocations in this file.

		The decoding of unwind sections for machine type Intel 80386 is not currently supported.

		Symbol table '.symtab' contains 2 entries:
		   Num:    Value  Size Type    Bind   Vis      Ndx Name
		     0: 00000000     0 NOTYPE  LOCAL  DEFAULT  UND
		     1: 00000000     0 NOTYPE  GLOBAL DEFAULT  UND _start

		No version information found in this file.

jackpot51:
That's interesting. I'll boot up my mac and see if I can take a look.

What does the readelf of build/terminal.rlib show?

esell:
$ greadelf -a terminal.rlib

        File: terminal.rlib(#1)
        readelf: Error: Not an ELF file - it has the wrong magic bytes at the start

        File: terminal.rlib(#1)
        readelf: Error: Not an ELF file - it has the wrong magic bytes at the start

        File: terminal.rlib(#1)
        readelf: Error: Not an ELF file - it has the wrong magic bytes at the start

        File: terminal.rlib(#1)
        readelf: Error: Not an ELF file - it has the wrong magic bytes at the start

        File: terminal.rlib(#1)
        readelf: Error: Not an ELF file - it has the wrong magic bytes at the start

        File: terminal.rlib(#1)
        readelf: Error: Not an ELF file - it has the wrong magic bytes at the start


I'll preface this by saying I have zero experience with rust but `make` does work in my ubuntu docker image. Everything comes out clean and I can copy over the disk image to my Mac and do a `make virtualbox` and it works. It's only crapping out when building locally on the Mac which is why I assumed it had something to do with the i386-elf buildchain I'm using. I spent awhile last night trying various flags without success.

jackpot51:
Thanks for putting in the work, it looks like rustc is not putting out the output specified in i386-elf-redox.json

I don't think it is a problem with the i386-elf tools from homebrew, I will see if rustc is doing the same thing on my system.

jackpot51:
I am getting travis-ci going, I will see if I can get a build going with the OS X VM, as I have a conflict between homebrew and macports.

jackpot51:
Found an issue with the target specification format. Can you do a git pull, make clean, and make virtualbox and confirm that virtualbox will run?

You may have to remove the vbox file in ~/VirtualBox VMs/Redox/Redox.vbox, I haven't fixed that yet.

esell:
Cleaned out all my old crap and did a fresh pull from git and everything seems to be working. Thanks for taking a look at! Looking forward to messing around with this.

PR #15, opened 2015-09-30
Add draft icons

Ticki:


Ticki:
The spacing between the electrons should probably be bigger. I'm gonna work on it tomorrow!

jackpot51:
It looks good, I set the redox-os orginazation's icon to it: https://github.com/redox-os

Issue #22, opened 2015-10-01
Filesystem


jackpot51:
This is a critical and massive piece of the system that has not been fully developed. At present, it has a single level node list, that contains the names and locations of every file. This entire list is loaded into memory when the system boots. The filesystem cannot be modified.

To fix this, there are a few options:
- Adopt another filesystem, and implement it.
- Fix the current filesystem by modifying the layout. This would likely involve:
 - Make the node list hierarchical
 - Add a free block list
 - Synchronize the filesystem with disk when a dirty flag is set
 - Online defragmentation

hauleth:
I would vote for using existing file systems. I would try to choose between modern FSes, my list:

- [Btrfs](https://en.wikipedia.org/wiki/Btrfs)
- [ext4](https://en.wikipedia.org/wiki/Ext4)
- [ZFS](https://en.wikipedia.org/wiki/ZFS)

All aof them are quite modern FSes and provide a lot of nice features. I would vote for ZFS as is IMHO the best option for the future.

stratact:
:+1: on ZFS support

jackpot51:
One important thing about ZFS is that it is CDDL licensed. This license is compatible with MIT, which is Redox's license, but not GPL.

BTRFS has better support on Linux, ZFS has better features (at the moment)

hauleth:
About Linux support I've created #25. As far as I can see Redox is MIT licensed and IMHO it is as it should be.

Also if we would implement ZFS from the ground then we can license it as we want. CDDL license is valid only for Sun implementation.

jackpot51:
I agree with the MIT license. We are all developing on Linux systems, and if ZFS-FUSE is good enough for us to create the boot disks in Linux, I would completely support implementing ZFS as the default filesystem in Redox.

hauleth:
More FSes could be added in the future, but IMHO ZFS as the first default FS would gather us a lot of attention which is what we need now.

jackpot51:
Choosing ZFS and moving discussion to #26

PR #24, opened 2015-10-01
Trim out more mut

stratact:
Please double check this change to see if this is a good alternative.


jackpot51:
It is good, take a look at #22 if you are active

PR #31, opened 2015-10-01
More vectors!

Ticki:


jackpot51:
I like it!

Ticki:
Thanks.

hauleth:
@Ticki you could crop SVG file to be squared and match icon size. It would look nicer.

PR #35, opened 2015-10-02
Broken writable filesystem

tedsta:
I did several things:
- Implemented `File::seek` and `File::flush` in libredox/file.rs
- Implemented `sys_lseek` and `sys_fsync` in syscall/call.rs
- Implemented `do_sys_fsync` and added it as handler for `SYS_FSYNC` in syscall/handle.rs
- Revised `Editor::save` in filesystem/apps/editor/editor.rs to write and flush the file.

If you edit a file, then save with F6, then reload with F5, you will see a lovely corrupted string.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@
 .DS_Store
 libc/build/
 setup/rust/
+*.swp
--- a/filesystem/apps/editor/editor.rs
+++ b/filesystem/apps/editor/editor.rs
@@ -37,6 +37,7 @@ impl Editor {
         let mut resource = File::open(&self.url);
         resource.seek(Seek::Start(0));
         resource.write(&self.string.to_utf8().as_slice());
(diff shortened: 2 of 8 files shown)


jackpot51:
Hey tedsta, I fixed it. The most major problem was in the file.rs scheme. On line 170, I added a check for a null extent (block or length are 0). On line 176, I converted the size of the extent into its size in sectors (it is stored in bytes, and it is confusing. It was done so that the exact length of a file could be determined)

https://github.com/redox-os/redox/blob/master/src/schemes/file.rs

jackpot51:
By the way, reallocation will be required soon, and it will need to rewrite the Node and also to reallocate the extents. We can talk about that tomorrow, if you want.

Thanks, and great job with the change!

tedsta:
Thanks! Sounds good.

PR #38, opened 2015-10-02
WIP: Prefix free function calls to their appropriate modules

stratact:
This is work-in-progress. Do not merge.

To help make code easier to navigate, I think it would be nice to have all of the free functional calls prefixed to their parent modules. This way you now exactly which precise functions are called and you can have multiple ones of the same name in the same module. In my opinion, this is a lot better than doing glob imports and it is more self-documenting in the source.

I would like people's opinions on this, before I continue on doing the same for rest of the free function calls. At the moment, I have only completed doing this with the `common::debug` module.


stratact:
I'm willing to take a :+1: or :-1: vote on this.

jackpot51:
In many cases, this is good. For debug, however, I chose very short function names because often dozens of things have to be outputted to the console, and I didn't want to type that much :-P At this point, many of those debug messages could be removed from the more stable modules, like ac97 and rtl8139. Thanks

filesystem/apps/editor/editor.rs.orig
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    fn save(&mut self, window: &mut Window) {
+        match self.file {
+            Option::Some(ref mut file) => {
+                window.title = "Editor (".to_string() + &self.url + ") Saved";
+                file.seek(Seek::Start(0));
+                file.write(&self.string.to_utf8().as_slice());
+                file.sync();
+<<<<<<< HEAD
```
hauleth: Some leftovers

jackpot51:
I cleaned up the leftovers in the merge/push to master. Thanks

PR #43, opened 2015-10-02
Big code style cleanup (almost all .rs files cleaned up)

Ticki:


jackpot51:
Ticki, can you get this to merge with master?

As a note, it is a very bad idea to continuously do style changes. I, for one, cannot follow all of these rules all the time, but that does not mean the code is unreadable.

Things I like:
- Spaces between function () and {
- Spacing out structs on one line per member

Some things I would not require as style:
- Commas after the last struct member? Many languages throw errors on this, like javascript, so it would be a very difficult rule to follow
- Putting function arguments on new lines. This should only be required if the function call is too long to read. In some places, this is absolutely unnecessary, like this code:
```
content.char(Point::new(8 * col, 16 * row),
                                 '_',
                                 Color::new(255, 255, 255));
```
Which, in my opinion, is better as this:
```
content.char(Point::new(8 * col, 16 * row),  '_', Color::new(255, 255, 255));
```

Many of these, then, should be developer decisions. We all have are own preferences, after all.

I definitely appreciate the cleanup, and early on, it is easier to take care of these issues. But if you have style items that will not be followed, I see it being a maintenance issue in the future.

Ticki:
> Putting function arguments on new lines. This should only be required if the function call is too long to read. In some places, this is absolutely unnecessary, like this code:

I get this one. However:

> Commas after the last struct member? Many languages throw errors on this, like javascript, so it would be a very difficult rule to follow

I agree, if I should make this choice about Rust I would disallow it, however this is the convention in Rust, and I personally think that we should follow that. Thoughts?

jackpot51:
If you get it to merge, this would change all of Redox to follow this convention, and I would start using a comma after the last struct member.

I would then be curious if we could count the number of violations of these concepts in a script. We could build a coding standard, and have a maximum line length to check for the function call seperation.

We could see how many violations of the comma rule there are and decide what to do then. We could also git blame the violations to see which contributors have which preferences.

What do you think? I like the cleanup, and think we need an automated way to check code style.

jackpot51:
I think a new directory, tests/, could contain pass/fail scripts for the build system to run

Ticki:
I could make a linter if you're interested.

jackpot51:
That would be good. I'm going to push the tests folder in with an example

tedsta:
Just thought I'd throw this out there: there is a code formatter for Rust in the works.

https://github.com/nrc/rustfmt

Not sure how usable it is yet.

Ticki:
Merge conflict almost resolved.

Ticki:
I'm giving up, I can't merge it.

jackpot51:
Ticki, don't give up! If you have it half way, make another pull request with the conflicted files checked in and I can help you merge!

jackpot51:
Hopefully, future style fixes won't have to touch as much code

Ticki:
Yay! It's seems to be solved!

jackpot51:
Good job Ticki!

Ticki:
Will push ASAP.

jackpot51:
Awesome, thanks

Ticki:
Done.

Ticki:
Should be mergable.

Ticki:
Yay! It merged!

PR #45, opened 2015-10-02
Write support

jackpot51:
This allows minor reallocation and messages when node is dirty or reallocation is not yet supported


jackpot51:
It runs correctly, and gives all the information required for future Node sector reallocation, so I will merge it into master

PR #52, opened 2015-10-03
Use cargo for `libredox`

Ticki:
TODO:
- [ ] Make src use cargo.
- [ ] Link them such that one can compile only using cargo (right now you need to use `make` to compile).


hauleth:
You can use [checklist](https://help.github.com/articles/writing-on-github/#task-lists).

Ticki:
Btw, this is safe to merge. It still compiles.

PR #56, opened 2015-10-04
Implement common::memory::Memory

stratact:
@jackpot51 does everything look good?

### DIFF
--- a/src/common/memory.rs
+++ b/src/common/memory.rs
@@ -21,6 +21,34 @@ struct MemoryMapEntry {
     acpi: u32,
 }
 
+struct Memory {
+    address: usize,
+}
+
+impl Memory {
+    pub fn new(size: usize) -> Option<Self> {
+        let alloc = alloc(size);
+        if alloc > 0 {
+            Some(Memory { address: alloc })
+        }
+        else { None }
+    }
+
+    pub fn renew(&self, size: usize) -> Option<Self> {
+        let realloc = realloc(self.address, size);
+        if realloc > 0 {
+            Some(Memory { address: realloc })
+        }
+        else { None }
+    }
+}
+
+impl Drop for Memory {
+    fn drop(&mut self) {
+        unalloc(self.address)
+    }
+}
+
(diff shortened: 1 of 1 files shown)


jackpot51:
It needs unsafe wrapping

src/common/memory.rs
```diff
@@ -21,6 +21,34 @@ struct MemoryMapEntry {
...
+    pub fn new(size: usize) -> Option<Self> {
+        let alloc = alloc(size);
+        if alloc > 0 {
+            Some(Memory { address: alloc })
+        }
+        else { None }
+    }
+
+    pub fn renew(&self, size: usize) -> Option<Self> {
+        let realloc = realloc(self.address, size);
```
jackpot51: should be let realloc = unsafe { realloc(...) };

src/common/memory.rs
```diff
@@ -21,6 +21,34 @@ struct MemoryMapEntry {
     acpi: u32,
 }
 
+struct Memory {
+    address: usize,
+}
+
+impl Memory {
+    pub fn new(size: usize) -> Option<Self> {
+        let alloc = alloc(size);
```
jackpot51: let alloc = unsafe { alloc(size) };

src/common/memory.rs
```diff
@@ -21,6 +21,34 @@ struct MemoryMapEntry {
...
+        if realloc > 0 {
+            Some(Memory { address: realloc })
+        }
+        else { None }
+    }
+}
+
+impl Drop for Memory {
+    fn drop(&mut self) {
+        unalloc(self.address)
```
jackpot51: unsafe{ unalloc(self.address) }

src/common/memory.rs
stratact: actually, I tried that, and rustc didn't like that

stratact: oh wait, nevermind. I'm stupid. What you suggest did work after trying it again.

jackpot51: Either way looks fine, now its time to propogate. I'm gonna take src/common/vec.rs

PR #57, opened 2015-10-04
Implement Memory::size() and readjust unsafe wrappers

stratact:


stratact:
@jackpot51 Did I do Memory::size() correctly?

jackpot51:
Yep, it is correct. :+1:

PR #58, opened 2015-10-04
Add vec!

jackpot51:


stratact:
@jackpot51 I gave it a shot in `common::context` by adding the `#![macro_use]` line at the very top of the module. I couldn't use `vec![]` where `Vec::new()` is since it still couldn't see the macro.

```
% make all
rustc --target=i686-unknown-redox-gnu.json -C no-vectorize-loops -C no-vectorize-slp -C no-stack-check -C opt-level=2 -Z no-landing-pads -A dead-code -A deprecated -L build -C lto -o build/kernel.rlib src/kernel.rs
src/common/context.rs:123:21: 123:24 error: macro undefined: 'vec!'
src/common/context.rs:123             memory: vec![],
                                              ^~~
error: aborting due to previous error
Makefile:148: recipe for target 'build/kernel.rlib' failed
make: *** [build/kernel.rlib] Error 101
```

stratact:
@jackpot51 

Okay, I think I figure out the madness with macros.

We need to put any macros in their own "macros" crate in order to use them outside of their scope, because `#[macro_use]` only works on `extern crate` lines, not `use foo::bar` lines, so putting them on `mod bar` inside existing parent module is useless for what we want.

Or heck, why don't we just include our own prelude that automatically `extern crate`s  the macros we want to use?

stratact:
@jackpot51 Any thoughts?

jackpot51:
Macros are weird, I will look at it when I get home

PR #59, opened 2015-10-04
update travisci to build on osx

esell:
Adds osx back in to travis-ci builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ before_install:
     - if [ `uname` = "Darwin" ]; then
         brew tap altkatz/homebrew-gcc_cross_compilers;
         brew update;
+        brew unlink gcc;
         brew install i386-elf-binutils nasm;
       fi
 script:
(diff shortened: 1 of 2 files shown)


esell:
Just a note that the osx build took about 18 minutes because of all the homebrew crap :(

esell:
I took out the gcc stuff, seems to work locally but the travis-ci stopped halfway through for some reason...

stratact:
@esell something about brew not linking

https://travis-ci.org/redox-os/redox/jobs/83516356#L2170

stratact:
Nice work!

esell:
ugh, missed a step, looks like it's good to go now!

hauleth:
ryan_reynolds_but_why

esell:
After deciphering your gif @hauleth I'm assuming you are asking why test the build on osx? I guess my thinking was since there are instructions in the readme on how to build on osx, we should probably make sure the build works.

For me it is quite frustrating to download a project and follow the documented steps to build it only to find out it doesn't work. You then spend 30 minutes trying to get the build working instead of playing around with/hacking on the project you just downloaded.

jackpot51:
Good job! I think we should test on OSX too

PR #68, opened 2015-10-05
[WIP] Window Management

stryan:
Current status of windowing system. Right now all I can think of for why it's not working is that it might be trying to open the url using a package instead of looking at Schemes.


jackpot51:
:+1: Getting started on this.

PR #69, opened 2015-10-05
Wman

jackpot51:


jackpot51:
@stryan I made filesystem/apps/player use the new window scheme. I will change some other apps, too

PR #70, opened 2015-10-05
Start ox

Ticki:
Ox can now:

- Download package data
- Parse .oxp files

TODO:
- Verify signatures
- Install packages


Ticki:
Btw, this PR also moves some files (to cargofy graphics module)

hauleth:
I would move Ox to it's own repository. Finally it shouldn't be tied to this one OS and should be external binary. Monorepos aren't good IMHO.

Ticki:
@hauleth Agree!

hauleth:
This will also simplify access management as team @redox-os/package-management could have RW access to external repo and they will manage and discuss their design independently from Redox.

jackpot51:
Creating ox repo, should it be redox-ox or ox? I think redox-os/ox would be good

hauleth:
I think that it should be `redox-os/oxide`.

jackpot51:
Thats good. Also, @hauleth did you get a slack invite?

hauleth:
Yes. I've joined seconds ago.

jackpot51:
Cool. The repo is up, hope @Ticki can make the first push and move the ox dir out of the kernel source

hauleth:
To help with that I suggest using `git subtree`. I can do that if @Ticki doesn't want to.

Ticki:
I will do ASAP.

PR #72, opened 2015-10-05
test window builds via appveyor

esell:
Looks like AppVeyor will do free builds for public repos. This PR adds in the appveyor.yml file to configure the builds and run them (it currently passes). If you want to use it you'll need to setup an AppVeyor account and connect it to the org.


jackpot51:
Cool! I will set it up

PR #81, opened 2015-10-06
Use `Self` where possible

stratact:
Makes it easier to maintain, if and when you change the name of the type.


stratact:
@jackpot51 All checks passed. Is there anything else you want me to cover before merging, chief?

PR #83, opened 2015-10-06
Refactor and Cleanup

stratact:
~~Work-in-progress. Don't merge yet.~~

Removing unnecessary `to_string()`s method calls and `String` arguments. This should hopefully speed things up a little with less allocation and more borrowing.

### DIFF
--- a/libredox/src/orbital.rs
+++ b/libredox/src/orbital.rs
@@ -21,13 +21,13 @@ pub struct Window {
 
 impl Window {
     /// Create a new window
-    pub fn new(x: isize, y: isize, w: usize, h: usize, title: &String) -> Self {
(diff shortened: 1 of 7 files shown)


libredox/src/orbital.rs
```diff
@@ -61,7 +61,7 @@ impl Window {
     }
 
     /// Set title
-    pub fn setTitle(&mut self, title: &String) {
+    pub fn setTitle(&mut self, title: &str) {
```
Ticki: Just a tiny nitpick: function should be in snake case.

stratact: oh, haha, good catch. Will take care of it.

stratact:
Up and ready to go!

@jackpot51 I find it better to use `format` if the first &str is going to be `to_string()`'d for concatenation. Better to clump them all together to be made in one `String`.

PR #89, opened 2015-10-07
Progress on nvpairs and XDR

tedsta:
Making some progress -- figured I'd make a PR now since it compiles at the moment. Don't want to fall too far behind master!


stratact:
Looks cool.

filesystem/apps/zfs/nvpair.rs
```diff
@@ -1,3 +1,25 @@
...
+
+// nvlist header
+pub struct NvList {
+    pub version: i32,
+    pub nvflag:  u32, // persistent flags
+    pub pairs: Vec<(String, NvValue)>,
+}
+
+impl NvList {
+    pub fn new(nvflag: u32) -> NvList {
```
stratact: I recommend changing the return from `-> NvList` to `-> Self` :sunglasses:

PR #122, opened 2015-10-10
Fix warnings

remram44:
This fixes warnings all over the code.

I left the warnings that only happen because of missing or temporary commented out sections.


filesystem/apps/test/test.rs
```diff
@@ -5,7 +5,7 @@ extern crate alloc;
 extern crate core;
 
 use alloc::boxed::Box;
-use std::{io, fs, rand};
```
stratact: Just so you know, there is code that is commented out that uses `std::fs` below

PR #133, opened 2015-10-14
&str == redox String cleanup

ambaxter:
I'd like SessionItem::scheme to return &str, too. However, that's a whole can of worms that ended up consuming me for far too long.


jackpot51:
:+1: Good job!

Issue #150, opened 2015-10-18
Getting started with redox-os


NobbZ:
Since I wanted to contribute somehow as a practical sideproject to the upcomming "building operating systems"-lectures this semester, I wanted to checkout redox-os, take a look at it, play around and perhaps contribute.

But one of the biggest problems is the following:

After building with ``windows/make all`` and then starting with ``windows/make virtualbox`` I do only see the following in my VirtualBox:

![redox-desktop](https://cloud.githubusercontent.com/assets/58951/10564511/a0dd685a-75b7-11e5-8741-beb12a610bd5.png)

When I hit CTRL-F1, I come back to the startup output and see the following content:

![redox-bootscreen](https://cloud.githubusercontent.com/assets/58951/10564515/bad973a2-75b7-11e5-8a51-6df95ad789fa.png)

From the first screen, I can move the mouse around, but nothing else happens, even after an hour of waiting. How can I get at least to the stage of a usable desktop as shown in the Screenshot from the README.md?

![https://github.com/redox-os/redox/raw/master/img/screenshots/Desktop.png](https://github.com/redox-os/redox/raw/master/img/screenshots/Desktop.png)

jackpot51:
Make sure your git newlines are \n and not \r\n

jackpot51:
See #111

NobbZ:
Yeah, following the advice in #111 I was able to compile and run. I hunted the issue down a bit and found about the possibility to configure this behaviour via ``.gitattributes``-file, so I sent a PR with an appropriate ``.gitattributes``.

PR #153, opened 2015-10-18
Added .gitattributes file

### LINKED ISSUE #111: No icons in Virtualbox on Windows 7
After transitioning to the GUI, Redox shows the wallpaper, but it doesn't populate the icons at the bottom of the screen as shown in the README.

I have tried 3 different versions of VirtualBox (4.2.? , 4.3.3, and finally, 5.0.6) and multiple revisions of the repo.

NobbZ:
This should preserve linux lineendings even under windows in all
circumstances.

This would also fix #111 and #150.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,2 @@
 # Make sure all Rust-Sourcefiles have Linux-style lineendings
-*.rs text eol=lf
+_REDOX text eol=lf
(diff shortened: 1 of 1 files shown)


jackpot51:
It is the _REDOX files that are the problem, not the rs files

NobbZ:
But perhaps it would be better not to rely on git preserving the lineendings but to remove that restriction and let redox repair them on the fly when the line ending is faulty.

jackpot51:
True, it would be better

PR #156, opened 2015-10-19
Implement file size and entry count for the file manager

stratact:
fm


stratact:
@jackpot51 :sunglasses:

PR #158, opened 2015-10-19
Fix Ordering for File Manager Sizes

Lacaranian:
- Ensure file manager checks sizes from large to small
-- Otherwise, MB/GB sizes would never be displayed


stratact:
@Lacaranian  I didn't know that if-else expressions have that kind of limitation where it needs to be order from large to small for intergers. Would it be better to pattern match these so that order doesn't matter?

PR #160, opened 2015-10-20
Make _REDOX parser more robust about lineendings

NobbZ:
This PR does make the parser for _REDOX-files more robust, by doing the following:

* Added a ``Lines``-Iterator for ``String``s
* Adding a ``lines``-function to ``String``
* Using these in ``kernel::programs::Package::from_url()``
* Reverting the quickfix from PR #153 

The added ``Lines``-Iterator should be robust enough to get along with all relevant lineendings that appear in the wild nowadays (Unix and Windows-Style), I don't think that we need to consider old Mac OS style (``\n`` only), there aren't even build instructions for such systems.

As well as PR #153 did, this PR addresses the (already closed) issues #111 and #150.


jackpot51:
Good job, it is exactly what was needed

PR #161, opened 2015-10-20
Added a batchfile for easier windows builds

NobbZ:
Altered README.md to reflect this.


jackpot51:
Interesting idea, I like it

PR #166, opened 2015-10-21
Implement Metric<T> enum for matching file sizes

stratact:
Because it's Rustic and cooler this way :sunglasses:


Lacaranian:
Since we're still using the if/else construct, we're only adding more code to do the same thing. Making this better might take more language features.

stratact:
Closing. Will come back to this in a new PR if we can find a way to not use the if/else construct and if necessary.

NobbZ:
I have some solution in my mind, I hope I will find some spare time...

stratact <notifications@github.com> schrieb am Mi., 21. Okt. 2015 19:18:

stratact:
@NobbZ cool and take your time.

PR #168, opened 2015-10-21
Lots of ZFS cleanup

tedsta:
- separated structs in zfs.rs to their own files
- created `ZFS::traverse` and use it to implement `read_file` and `ls` to reduce duplicated code
- added comments


jackpot51:
Very cool!

PR #173, opened 2015-10-23
Implement `FileType` + Use hashmap to hold file info + Add file description in file manager + Cleanup

stratact:
This is not complete, because when I open `file:///ui/`, the file manager is not happy with the full window width.

problem


jackpot51:
I don't see the problem immediately, but it was there before the descriptions (with just the sizes)

PR #176, opened 2015-10-23
Fix annoying event bug in Sodium

Ticki:
Thanks for the help, Stratact and Tennix!


stratact:
Well, you're the one who really found it, all I was able to do is narrow it down, but I didn't realize it was the events causing it. :smile:

PR #177, opened 2015-10-24
move color struct to userspace, swap it in for [u8] all over the place

LazyOxen:
Made the PR rhyme, so you can't deny it.
Oh yeah, also fixed the variable naming in bmp.rs - the red and blue channels were swapped


ambaxter:
+1 Pretty straight forward.

PR #198, opened 2015-10-28
Add QuitEvent

nounoursheureux:
This provides an interface for the window manager to ask for a window to exit, without directly killing the window. 
If the idea is accepted I will handle this event in all the already written applications.


jackpot51:
Very useful, thanks!

PR #200, opened 2015-10-28
Make URL syntax agnostic, rewrite path parser (is about 3 times faster), rewrite IP parser, improve performance in kernel, add additional debug messages, provide default methods for Resource trait, utilize &str over String in a lot of places, avoid problems in the filesystem related to file naming (parsing), remove return keywords, rustify code, clean up code style, remove a bunch of warnings, remove unused imports, refactor and move around things in the kernel, refactor things in Sodium, improve performance in Sodium.

Ticki:
Fixes a bunch of issues too.


Ticki:
It seems that Redox is now about 10-20% faster when booting.

PR #202, opened 2015-10-28
Have to explicitly match or will still `1 + string.len()`

stratact:
Apparently `unwrap_or` is not great for everything.


stratact:
@Ticki we really need that `get_slice()` method

Ticki:
Will work on it tomorrow.

PR #203, opened 2015-10-29
Cleanup in `kernel::common::parse_ip`

stratact:


stratact:
@jackpot51 is this mergeable?

kernel/common/parse_ip.rs
```diff
@@ -1,37 +1,36 @@
...
-            _ => b += 1,
-        }
-    }
-
-    &string[..b + 1]
-
+    let pos = match string.find(':') {
+        Some(pos) => pos + 1,
+        None => match string.find('/') {
+            Some(pos) => pos + 1,
```
Ticki: That's quite inefficient, you make two searches in the string.

stratact: Alright, but is there a way to say, if the loop finishes we wouldn't be doing a `&string[..b + 1]`? That could mean a possible panic if `:` or `/` were not found.

stratact: Or we should just probably wait for that `get_slice()` method before doing anything.

Ticki: Shall I optimize it?

stratact: Sure, just keep in mind to account for the possibility that if it can't find those two chars, the slice should be `&string[..b]`, not `&string[..b + 1]`

PR #209, opened 2015-10-29
Add debug! and debugln! macros for debug syscalls (to print to system console)

Ticki:
I thought this was needed for debugging.

### DIFF
--- a/libredox/src/console.rs
+++ b/libredox/src/console.rs
@@ -73,6 +73,24 @@ macro_rules! println {
     });
 }
 
+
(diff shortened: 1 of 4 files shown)


jackpot51:
Thanks, really needed these!

stratact:
Nice work!

jackpot51:
Can you add these to the kernel as well?

do_sys_debug is the kernel equivalent to sys_debug

Ticki:
Will do.

PR #215, opened 2015-10-31
Fix bug in sodium + Add line marker

Ticki:
:tada: I found the :bug:


stratact:
Congrats on killing the bug 
:bug: :wavy_dash: :gun:

Ticki:
:fire: :bug: :fire:

PR #217, opened 2015-10-31
Make division and modulo in the kernel 2-5 times faster. Sodium: update statusbar, add scrolling commands, add `get` command

Ticki:


Ticki:
Oh forgot to add a file. Now it's fixed.

PR #220, opened 2015-11-02
Implement get_slice() for bounded slice abstraction

stratact:
Implemented the long awaited get_slice() method for both `&str` and `&[T]` :tada:


Ticki:
Good work.

PR #222, opened 2015-11-02
Adjust GetSlice doc comments

stratact:


stratact:
@Ticki thanks for the help
@jackpot51 the comments do what I want now.

PR #224, opened 2015-11-02
ZFS - dnode::ObjectType enum, comments, more space map stuff

tedsta:


jackpot51:
So close to new allocations, good job!

PR #226, opened 2015-11-03
Adjust the type name acronym casing to fit the Rust convention

### LINKED ISSUE #61: Follow rust naming convention
BMP should be Bmp
WAV shoud be Wav

and so on.

stratact:
Closes #61


stratact:
Also I left `filesystem/schemes/zfs/` alone, because the `filesystem/apps/zfs/` code is more updated and current for libzfs.

PR #247, opened 2015-11-06
Remove WM from the kernel space. Redox now boots to console. Clean up a bunch of code in kernel space. Remove the session module from kernel space.

Ticki:
It compiles.


Ticki:
Note that it still needs loooooooots of cleanups and uncommenting stuff which is needed, however it compiles and boots just fine.

nounoursheureux:
Great ! I really thought that it was weird to have the WM in kernel space, but unfortunately I wasn't familiar enough with the codebase to make this change myself.

stratact:
@Ticki great work

@nounoursheureux there was a lot of stuff that needed to be moved from the kernel, that was supposed to be in user space.

PR #252, opened 2015-11-06
Make paths relative (utilize the new parser)

Ticki:


jackpot51:
This is not exactly what I visualized, it needs to be called in Context::canonicalize.

The reasons are that accessing the cwd in every scheme is ugly, especially considering the overhead in userspace schemes.

So, canonicalize happens once on every path passed to open() or exec() so that the schemes are always given absolute paths

PR #256, opened 2015-11-08
Update README.md

k0pernicus:
Extra information about "which version of Rust to use?"


Ticki:
Thanks. This was needed.

PR #263, opened 2015-11-10
Get the current directory in the prompt terminal

k0pernicus:
New feature : the current directory in the prompt terminal!


jackpot51:
Very nice!

PR #268, opened 2015-11-10
Rename `Arc` to `ArCache` in ZFS

stratact:


tedsta:
Just want to say for the record that this is because of the naming conflict with `std::sync::Arc`

PR #271, opened 2015-11-11
File manager

LazyOxen:
Fixed: Navigation works again
Functional changes: file_manager only uses one window instead of multiple windows
Quirks: changing directories moves the window, since there's no resizing yet


jackpot51:
Nice job! I think the window movement can be fixed by passing the old window x, y to the constructor. It just isn't hooked up yet

PR #278, opened 2015-11-13
Usual cleanup

stratact:


liborbital/bmp.rs
```diff
@@ -135,11 +136,6 @@ impl BmpFile {
         !self.data.is_empty()
     }
 
-    /// Convert to slice for drawing
-    pub fn as_slice(&self) -> &[Color] {
-        &self.data
```
Ticki: Why not use the actual deref coercion here?

stratact: Well, first, what's the point of keeping `as_slice` since `&bmp` does the same thing and it's shorter?

Ticki:
:+1:

PR #283, opened 2015-11-14
Navigate to parent folder. Fix file information error I introduced.

LazyOxen:
Quirks: shows parent directory at the top level.


jackpot51:
Very good stuff!

PR #298, opened 2015-11-20
Cleanup GetSlice

stratact:


stratact:
Our libcore needs to be updated.

Ticki:
I'll update the libs.

PR #302, opened 2015-11-20
Fix bugs in malloc

Ticki:
Dont code at late night.


Ticki:
Wow, this code is buggy. Hopefully many of them are solved now. It still wont boot, though.

PR #305, opened 2015-11-21
Introduce SliceData trait for generic types representing slices

stratact:


stratact:
@jackpot51 `Vec<T>` and `String` are covered, since they `Deref` down to `[T]` and `str` respectively anyway.

PR #311, opened 2015-11-22
Few changes & clean some useless code in 'shell'

k0pernicus:


jackpot51:
Read command is for reading variables, run reads and runs a script file, ypur change would break that

k0pernicus:
Oh ok, sorry to hear that.
I will update the documentation so. Thanks!

PR #315, opened 2015-11-24
Real hardware

jackpot51:
Lots of process management changes
Lots of kernel cleanup
Still has a bug in the Terminal application (when exec is used)


Ticki:
:tada:

PR #318, opened 2015-11-26
Fix bugs in malloc's tree structure related to bit manipulation

Ticki:


Ticki:
Now it allocates! But deallocation segfaults.

PR #332, opened 2015-11-29
Add git submodule steps to README

skylerberg:
A recent change made the README setup instructions fail because rust is
not pulled as a submodule. This commit adds an additional step to the
setup documentation for each operating system indicating what to do to
fetch the submodule for rust. This change has only been tested on
Ubuntu.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,13 @@ Sometimes things goes wrong when compiling. Try the following before opening an 
 
 and then rebuild!
 
+## Cloning the repository
+
+Make sure you get submodules when you clone the repository.
+```bash
+git clone --recursive
+```
+
 ## Building on Debian/Ubuntu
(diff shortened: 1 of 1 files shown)


hauleth:
Instead one should use `git clone --recursive`.

skylerberg:
@hauleth Thanks! I have updated the commit to use `git clone --recursive`.

polymetric1:
Problem with this is, I already had the repo saved, so when my build failed, I'd have to reclone the repo. I didn't, because I looked it up and you simply have to run `git submodule update --init --recursive`. I think this should be mentioned in the README

Ticki:
You actually only need `git submodule update --init`.

skylerberg:
@polymetric1 I will add another commit to this branch covering that.

skylerberg:
Pull request created with updated instructions #333.

PR #334, opened 2015-11-30
Finish TODO for MFU in ZFS

stratact:
@tedsta

### DIFF
--- a/filesystem/apps/zfs/arcache.rs
+++ b/filesystem/apps/zfs/arcache.rs
@@ -22,7 +22,7 @@ impl Mru {
         }
     }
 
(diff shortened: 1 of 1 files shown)


filesystem/apps/zfs/arcache.rs
```diff
@@ -61,11 +61,26 @@ impl Mfu {
...
+                    lowest_dva = Ok(dva_key);
+                }
+            }
+
+            self.map.remove(&try!(lowest_dva));
+        }
+
+        // Add the block to the cache
+        self.used += dva.asize() as usize;
+        self.map.insert(*dva, (1, block));
```
stratact: @tedsta I had the frequency start at `1`, but should it start at `0` when caching?

PR #335, opened 2015-12-01
Code cleanup

stratact:
In the shell:
* Described what the `help` argument is used for in the doc comments
* Reduce lines of code by idiomizing in Rust more
* Removed unneeded cloning, because cloning a `&str` is not necessary to use `to_string()`.


stratact:
In ZFS:
* Removed the prelude imports which are already imported
* Cluster multiple import lines into one line
* Unglob imports with explicit ones

PR #341, opened 2015-12-02
Make use of fewer raw pointers, use borrows instead.

Ticki:
Note that this do not have performance implication since Option<&T> is optimized to a nullable pointer. As such, this change only affects safety.

It cannot panic (even though it uses `.unwrap()`), because the `None` case is unreachable (and is even optimized away if you look at the LLVM IR output).


LazyOxen:
Neat! :+1:

PR #347, opened 2015-12-05
Fix alloc_align and fix address map

Ticki:
Currently broken.


Ticki:
For some reason this boots to a black screen even though it works fine at my local machine (not in Redox).

Issue #345, opened 2015-12-05
Orbital page faults after exec fails in terminal


skylerberg:
Steps to reproduce:

* Switch to desktop
* Open terminal
* Run the command `exec foo`
* Wait 2-3 seconds

A couple seconds after exec foo fails orbital will page fault and stop working.

Here is the output when orbital fails:
```
file:/foo: Failed to open
PID 7: file:/schemes/orbital/main.bin
  INT E: Page fault
    CS:  00000008    IP:  C0004185    FLG: 00010046
    SS:  00000000    SP:  C0015000    BP:  0180FEBC
    AX:  C149A000    BX:  00000001    CX:  00000001    DX:  00000002
    DI:  00000000    SI:  00000000
    CR0: 80000013    CR2: 00000029    CR3: 00200000    CR4: 00000600
    ERR: 00000002
```

polymetric1:
Didn't happen for me. I repeated it the exact same way. Ran `exec foo` from orbital desktop. Can you maybe give a bit more info?
- What VM are you using? Try using the alternative, just to see if that works.
- Have you updated your VM lately?
- What is your host OS?
- Are you building the buddy branch or master? If you're using buddy, try master.
It's probably a problem with qemu/virtualbox.

skylerberg:
Yeah, I am running Ubuntu 14.04 using qemu. Virtualbox is not working on my host at the moment, so I cannot test it. I am running on master. qemu is up to date on the system.

Ticki:
I ran into page faults multiple times (when debugging Sodium) however I'm unable to reproduce.

skylerberg:
I just realized it doesn't just page fault after a bit. I have to do something on orbital. For example, move my cursor around for a bit or minimize and maximize the window a couple times.

Ticki:
Are you in sync with master?

skylerberg:
Yes, I am in sync with upstream master.

Ticki:
Orbital needed some heavy refactoring anyways.

polymetric1:
Virtualbox is not working for me, either. Same version of Ubuntu. However my setup is Redox(qemu) running on Ubuntu(virtualbox) running on Windows(host), which I feel like would cause more errors.
It's not a problem with qemu, then, anyway. Page faults seem to occur rather commonly, but are not easily reproducable and seem random. I just got one simply by opening the terminal in orbital.

Ticki:
[...]

PR #348, opened 2015-12-05
Modify Command::spawn to check for file existance

LazyOxen:
Addresses #345 
Just a thought: Would it be worth considering making system calls work on file handles instead of strings to avoid file access race conditions?


LazyOxen:
Unsatisfactory "solution".

PR #356, opened 2015-12-08
Run `rustfmt` on `kernel`, `libredox` and `liborbital`

polymetric1:
Note: I used a cutting edge debug version of `rustfmt`, so if anyone sees any problems, say something immediately, don't hesitate, even if you don't think it's a problem. I did build, run and test this though, and the only functional problems this could introduce would be rejected by the compiler, so I think it worked fine.

### DIFF
--- a/libredox/src/url.rs
+++ b/libredox/src/url.rs
@@ -4,15 +4,15 @@ use core::mem;
 use string::{String, ToString};
 use vec::Vec;
 
-//URL Parsing:
-//Split by /
-//scheme://user:password@host:port/path/path/path?query#fragment
-//First part is scheme, second is empty, third is user, password, host, and port, later parts are path, last part is path, query, and fragment
-    //Split third part by @, the last part is the host and port, if there is a first part it is the user and password
-        //Split these parts each by :, first part splits into user and password, the second part is split into domain and port
-    //Split the last part by ?, the first part is a path element, the last part is the query and fragment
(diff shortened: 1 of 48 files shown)


libredox/src/url.rs
```diff
@@ -4,15 +4,15 @@ use core::mem;
...
-            //Split the query by &
+// URL Parsing:
+// Split by /
+// scheme://user:password@host:port/path/path/path?query#fragment
+// First part is scheme, second is empty, third is user, password, host, and port, later parts are path, last part is path, query, and fragment
+// Split third part by @, the last part is the host and port, if there is a first part it is the user and password
+// Split these parts each by :, first part splits into user and password, the second part is split into domain and port
+// Split the last part by ?, the first part is a path element, the last part is the query and fragment
+// Split the last part by #, the first is the query, the second is the fragment
+// Split the query by &
```
stratact: From lines 11-15, it appears they were indented on purpose to show flow of instructions. However instead of indenting the lines themselves, could you add indentation between the `//` and `Split` in the same style as the diff?

stratact: Or instead of using spaces, how about using `*` as bullet points, like in markdown?

polymetric1: Good point. I'll try to fix that

polymetric1:
I'll redo this and create a new PR later.

PR #360, opened 2015-12-09
Fix some problems in CONTRIBUTING.md

polymetric1:


Ticki:
Other things:

- Follow the style conventions
- Use `.into()` and `.to_owned()` over `.to_string()`.
- Prefer passing references to the data over owned data. (Don't take `String`, take `&str`. Don't take `Vec<T>` take `&[T]`).
- Use generics, traits, and other abstractions rust provides.
- Make sure to mark parts that needs work on with `TODO`, `FIXME`, `BUG`, and `UNOPTIMIZED`.

PR #365, opened 2015-12-11
added support for german keyboard layout

fischmax:
I cannot test this though, since the build currently fails. The next step would be to support additional modifier keys to be able to insert [ ] { }. If this gets accepted, I could delve into that task ;-)


Ticki:
Cool stuff. There reason it fails with travis is that redox-os/rust is not even with upstream master rust. I've send a PR that fixes that :)

Ticki:
@jackpot51 This patch is working now. Can you merge it?

PR #372, opened 2015-12-15
Update README

k0pernicus:
Contents added & refactoring of the Markdown


polymetric1:
Nice work, you think you could do something like this on CONTRIBUTING.md? I'd do it but I'm really not good at formatting and refactoring markdown like this

k0pernicus:
No problem! :-)

PR #379, opened 2015-12-16
Add support for right alt key (altgr)

fischmax:
I currently don't have access to english and french keyboards, so the altgr-mappings for these layouts might be flawed. But this can easily be fixed afterwards.


polymetric1:
Suggestion: You're doing alot of merging, try rebasing instead (`git pull --rebase upstream master`).

fischmax:
Yeah, At my  current workplace, we aren't allowed to rebase, so I kinda lost the habit. I'll try to remember it in the future.

Ticki:
@polymetric1 Why don't merge? It gives better history than rebasing.

polymetric1:
@Ticki The problem with this is that it creates useless commits that are added to pull requests that do nothing. It's fine for _merging_ branch-x with the master branch, hence the name, but when you're syncing a branch with the master, it's better to 1. just don't, or 2. rebase, than to merge. You shouldn't always merge and you shouldn't always rebase. This is kind of a bad explanation but I hope you understand what i mean

jackpot51:
I cannot merge this because it backrevs the rust submodule.

You will need to ```
cd rust
git checkout master
git pull
```

And resubmit

polymetric1:
Merging also causes problems like this ^

stratact:
http://www.derekgourlay.com/archives/428

polymetric1:
- Rebase when pulling upstream changes
- Merge when finishing a separate feature branch
Couldn't have said it better.
Actually I tried to explain it, I'm just that bad at explaining things.

Ticki:
> Rebase when pulling upstream changes

Of course.

PR #383, opened 2015-12-17
added bootstrap file

DomThePorcupine:
Hey guys this is my bootstrap file, I have not been able to test it for Fedora or SUSE but everything should work. I was thinking this would make things a little easier, especially with the readme because it could be shortened to something like 
`http://github.com/rawfile/bootstrap.sh | bash`
instead of all those instructions for the individual distros.
I am not familiar at all with NIXos so if someone else knows about that it would be great.

The script also has a useful update function:
`./bootstrap -u`
which will update the rust compiler with the binary.sh function and also pull the newest version of redox

This script will also allow you to remove a bunch of the specific distro.sh files(except arch although it wouldn't be that hard to add it).

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,0 +1,175 @@
+#! /bin/bash
+banner() {
+	echo "|------------------------------------------|"
+	echo "|----- Welcome to the redox bootstrap -----|"
+	echo "|------------------------------------------|"
+}
+
+osx()
+{
+	echo "Detected OSX!"
+	temp=$(brew --version)
+	if [ -n "$temp" ]; then
+		echo "Homebrew detected! Now updating..."
+		brew update
+		echo "Now installing git..."
+		brew install git
+		echo "Now installing virtualbox..."
+		brew cask install virtualbox
+	else
+		echo "Homebrew does not appear to be installed! Would you like me to install it?"
+		printf "(Y/n): "
+		#For now assume yes
+		ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+	fi
+	echo "Cloning Redox repo"
+	git clone -b $1 --recursive https://github.com/redox-os/redox.git
+	sh redox/setup/osx-homebrew.sh
+	sh redox/setup/binary.sh
+}
+
+archLinux()
+{
+	echo "Detected Arch Linux, mah fav"
+	echo "Updating system..."
+	sudo pacman -Syu
+	echo "Installing git..."
+	sudo pacman -S git
+	echo "Cloning Redox repo"
+    if [ "$2" == "qemu" ]; then
+        echo "Installing QEMU..."
+        sudo pacman -S qemu
+    fi
+	git clone -b $1 --recursive https://github.com/redox-os/redox.git
+	sh redox/setup/arch.sh
+	sh redox/setup/binary.sh
+}
+
+ubuntu()
+{
+	echo "Detected Ubuntu/Debian"
+	echo "Updating system..."
+	sudo apt-get upgrade
+	echo "Installing git..."
+	sudo apt-get -y install git
+    if [ "$2" == "qemu" ]; then
+        echo "Installing QEMU..."
+        sudo apt-get -y install qemu-system-x86 qemu-kvm
+    fi
+	echo "Cloning Redox repo"
+	git clone -b $1 --recursive https://github.com/redox-os/redox.git
+    sudo apt-get -y install build-essential libc6-dev-i386 nasm curl file
+    sh redox/setup/binary.sh
+}
+
+fedora()
+{
+    echo "Detected Fedora"
+    echo "Installing git..."
+    sudo yum install git-all
+    if [ "$2" == "qemu" ]; then
+        echo "Installing QEMU..."
+        sudo yum install qemu-system-x86 qemu-kvm
+    else
+        echo "Installing virtualbox..."
+        sudo yum install virtualbox
+    fi
+    echo "Cloning Redox repo"
+    git clone -b $1 --recursive https://github.com/redox-os/redox.git
+    sudo dnf install gcc gcc-c++ glibc-devel.i686 nasm make
+    sh redox/setup/binary.sh
+}
+
+suse()
+{
+	echo "Detected a suse"
+	echo "Installing git..."
+	zypper install git
+	if [ "$2" == "qemu" ]; then
+		echo "Installing QEMU..."
+		sudo zypper install qemu-x86 qemu-kvm
+	else
+		which virtualbox && { "Virtualbox installed!"; break; }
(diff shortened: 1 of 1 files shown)


bootstrap.sh
```diff
@@ -0,0 +1,175 @@
...
+	git clone -b $1 --recursive https://github.com/redox-os/redox.git
+	sh redox/setup/arch.sh
+	sh redox/setup/binary.sh
+}
+
+ubuntu()
+{
+	echo "Detected Ubuntu/Debian"
+	echo "Updating system..."
+	sudo apt-get upgrade
```
NobbZ: 1. This does only work if the user has made a manual `apt-get update` before. `upgrade` does not enforce it.
2. Also as well as `install``update` needs manual confirmation of the packages that shall be upgraded.
3. There are situations when one does not want to mass-upgrade his system. eg. during the lecture-phase I am very picky with upgrades, since some upgrades could destroy stuff I have to rely on.

bootstrap.sh
```diff
@@ -0,0 +1,175 @@
...
+	echo "Installing git..."
+	sudo apt-get -y install git
+    if [ "$2" == "qemu" ]; then
+        echo "Installing QEMU..."
+        sudo apt-get -y install qemu-system-x86 qemu-kvm
+    fi
+	echo "Cloning Redox repo"
+	git clone -b $1 --recursive https://github.com/redox-os/redox.git
+    sudo apt-get -y install build-essential libc6-dev-i386 nasm curl file
```
NobbZ: You are missing a new status-message here. This is not cloning the repository anymore. Do them right or leave them out.
Also I dont like that `-y`, since it does not give me a choice. Also multiple runs of `apt` are enforced, why dont build up a list of packages and then install all in one run letting the user choose if he really want to do by ommiting `-y`.?

Some of the comments might apply for other OS as well

NobbZ: `apt` is not the sole package manager for Ubuntu, there is also `aptitude` and `apt-fast` (OK, it is a wrapper for `apt` and `aptitude` that predownloads the packages in parallel), I'd like to choose either by commandline argument or by environment variables.

NobbZ:
I have left some line comments,

Also I don't like such scripts. A README or any other textfile is easily scannable for the headlines while such scripts are rarely that easy to read as this one, also it might grow and become much less tidy. Then it is hard to find the relevant parts for my OS.

I do not trust arbitrary installers that do a `curl $URL | *sh` or similar…

If you really want to clean up the README, pull that stuff out into some external build documentation.

bootstrap.sh
```diff
@@ -0,0 +1,175 @@
...
+	echo "Updating system..."
+	sudo apt-get upgrade
+	echo "Installing git..."
+	sudo apt-get -y install git
+    if [ "$2" == "qemu" ]; then
+        echo "Installing QEMU..."
+        sudo apt-get -y install qemu-system-x86 qemu-kvm
+    fi
+	echo "Cloning Redox repo"
+	git clone -b $1 --recursive https://github.com/redox-os/redox.git
```
NobbZ: You do assume that I do not already have the sources and pull it straight away into the folder which is my current working dir. Perhaps It is already exactly here, perhaps I want it to be somewhere else?

When I really do run such installers, I do it **always** from inside a temporary location, then it is easier to clean up the mess that some of them leave around…

bootstrap.sh
DomThePorcupine: So would you suggest removing the update part completely?

bootstrap.sh
DomThePorcupine: @NobbZ 
Do aptitude/apt-fast work the same as apt-get? For example would 'apt-get install package-x' be the same as 'aptitude install package-x' and 'apt-fast install package-x'? I do not work in Ubuntu often, so I am not super familiar with these alternate package managers.

bootstrap.sh
NobbZ: Since an upgrade of the system is not really necessary to get up and
running redox I think, drop it. Still we need to update package repositorys
via `apt-get update` to avoid problems when installing later. If you don't
do it, you might run into troubles because package versions shifted since
last update.

DomThePorcupine <notifications@github.com> schrieb am Do., 17. Dez. 2015 um
08:01 Uhr:

bootstrap.sh
NobbZ: The interface is pretty much the same, at least the relevant parts…

The sole difference in the interface is how upgrades happen. There are `upgrade` and `dist-upgrade` in `apt` while there are `safe-upgrade` and `full-upgrade` in `aptitude`. If we remove upgrading, it would be irrelevant.

`apt-fast` just mimics the interface of the tool it is wrapping.

bootstrap.sh
DomThePorcupine: @NobbZ 
So would you suggest requiring an directory location given by the user, something like
"Put repo in CWD?"
"(Y/n):"
if n
"Please enter full path to location:"

bootstrap.sh
DomThePorcupine: @NobbZ 
So drop `apt-get upgrade` in favor of `apt-get update`?

bootstrap.sh
DomThePorcupine: @NobbZ 
Okay so a command line option for this would not be difficult to implement at all then!
Also to address the -y, I attempted to run the script in a hastily prepared chroot and running the script it did not wait for me to press y it just assumed n and exited so that is why they are there. I guess I messed something up with that, I will drop them.

polymetric1:
This is a pretty good shell script, I just don't think it really follows convention, as you usually will want to do all this manually, because of the fact that this script is part of the repository and will be always updating, there is always potential for problems. You could have the script pull from origin, or some other screwup that just, isn'f necessary for the sake of slight convienence. This isn't even like the code of conduct thing, I've never seen ANY project with this, ever. Then again, I could say the same for Redox's setup and windows folders, but honestly this is a bit too far. Whenever I clone Redox onto a new machine, I execute every command individually. I'm sure many others can say the same. You say you're new to Git and GitHub, once you're more experienced you'll see how this script really isn't necessary and can just build up problems. I'm kinda against having this script in the main repo. Feel free to have your own and .gitignore_global it or something maybe. Mostly just my opinion though.

Ticki:
@polymetric1 I do not agree. This script is greats for newbies, who want to get start and hack on Redox. I think this great script should be added.

jackpot51:
@polymetric1 it is very similar to rustup

polymetric1:
@Ticki @jackpot51 yeah, I think I got a little ahead of myself, yeah I guess the script is fine. I just overthibk things like this, but I actually really like rustup so I think it is a good idea. Some things can be changed as always, though.

Ticki:
Exactly. These kind of scripts are really great, since people often don't want to invest much time to get started. I think this might attract lots of new users/hackers.

PR #389, opened 2015-12-20
Don't reexport everything in the root of libredox + Use less pointer sized integers (*size -> *32 in liborb{ital, tk})

Ticki:


Ticki:
This patch is now fixed.

filesystem/apps/editor/main.rs
```diff
@@ -1,20 +1,21 @@
...
+use std::env;
 
 pub struct Editor {
     url: String,
     file: Option<File>,
     string: String,
-    offset: usize,
-    scroll_x: isize,
-    scroll_y: isize,
+    offset: u32,
```
jackpot51: Why make offset u32, if it significantly increases the number of `offset as usize` statements?

filesystem/apps/terminal/window.rs
```diff
@@ -19,20 +18,20 @@ pub struct ConsoleWindow {
     /// Previous commands
     pub history: Vec<String>,
     /// History index
-    pub history_i: usize,
+    pub history_i: u32,
     /// Offset
-    pub offset: usize,
+    pub offset: u32,
```
jackpot51: Again, I don't want to see offset be `u32` because of the number of awkward, unnecessary conversions

liborbtk
```diff
@@ -1 +1 @@
-Subproject commit 18573e091d9dacd466940ea97cb45722bc89956d
+Subproject commit a43b52420b0bdc3ad292062415b7212a901711dd
```
jackpot51: This submodule was reverted

libredox/src/to_num.rs
```diff
@@ -4,14 +4,14 @@ use get_slice::GetSlice;
 
 /// Parse the string to a integer using a given radix
 pub trait ToNum {
-    fn to_num_radix(&self, radix: usize) -> usize;
-    fn to_num_radix_signed(&self, radix: usize) -> isize;
-    fn to_num(&self) -> usize;
-    fn to_num_signed(&self) -> isize;
+    fn to_num_radix(&self, radix: u32) -> u32;
```
jackpot51: Perhaps to_num<T> makes more sense than forcing u32 here? It would limit parsing to numbers below 4 billion

jackpot51:
I like that all points are i32, all sizes are u32. I support merging in these parts

I very much dislike using u32 for an offset, offsets are usize because they are defined in pointer arithmetic types. Therefore, you have dozens of casts from u32 to who-knows-what-size usize when you could easily specify usize for all offsets and remove these casts

I would prefer u64 and i64 for ToNum, because it is generally better to accept more user input and then let the program decide how to clip it to a smaller number, if desired

Ticki:
> I would prefer u64 and i64 for ToNum, because it is generally better to accept more user input and then let the program decide how to clip it to a smaller number, if desired

I disagree, lossless conversions are always better.

jackpot51:
Then you should have the functions be `to_num<T>`, which I also like

Ticki:
Yeah, that's right, it should be generic.

Ticki:
I just did some changes to CONTRIBUTING.md too, cc @polymetric1

Ticki:
I'll fix your points, @jackpot51

Ticki:
@jackpot51 I believe it's fixed now.

Ticki:
Yeah, sorry.

Ticki:
Needa fix that tonum trait.

jackpot51:
I appreciate it @Ticki

PR #390, opened 2015-12-21
Organize README.md and CONTRIBUTING.md

polymetric1:
I removed the inconsistent and unorganized multiplatform build instructions in favor of two instruction sets, quick setup and manual setup. Quick setup uses `bootstrap.sh`, and manual setup just lists all the commands you need to do it manually. This should make it alot more simple and easy for beginners, as well as more advanced users.

The diff is a bit hard to read, you can just look at the final changed files here https://github.com/polymetric1/redox/blob/contributing-md/README.md
https://github.com/polymetric1/redox/blob/contributing-md/CONTRIBUTING.md

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -3,7 +3,7 @@
 Thank you for your interest in contributing to Redox! This document is a guide to help newcomers contribute!  
 There are many ways to help us out and we appreciate all of them.
 
-### Index
+## Index
 
 * [Communication](#communication)
  * [Slack Chat](#slack)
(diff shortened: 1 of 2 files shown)


DomThePorcupine:
I think that when you pipe this directly into sh it doesn't wait for you to confirm installs. So instead of installing the packages it will just skip over them. I had flags in for automatic installation of the packages in ubuntu but @NobbZ thought that that was a bad idea.

polymetric1:
@DomThePorcupine Yeah, your bootstrap script is a really great idea but I think it needs some major refactoring before it can be used. I might split this into two pull requests, one for just the Quick Setup section until we get that refactored.

DomThePorcupine:
@polymetric1 There are some fairly major changes waiting in my fork, I was about to open up a new pull request. I will do that soon.

polymetric1:
Alright. I had some changes in mind, I'll wait for yours to be pulled before I update the script and the README

DomThePorcupine:
@polymetric1 The script should automatically delete the bootstrap file when done so the `&& rm bootstrap.sh` will not be necessary

polymetric1:
Ah ok. I knew I was going to far with that.

PR #391, opened 2015-12-21
Fix bootstrap.sh problems

DomThePorcupine:
There were a lot of problems pointed out to me with the initial pull of this file. I think I have fixed them all,
* Fixed allowing for alternative package managers under Debian/Ubuntu
* Fixed assuming yes to installing packages, now user is prompted to agree
* Fixed blindly installing git, QEMU, Virtualbox, etc. these things are tested for and skipped over if installed
* Changed default emulator to QEMU
* Added ability to install QEMU on OSX
* Added comment advising `qemu_no_kvm` on OSX
* Cleaned up other general things

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -36,6 +36,7 @@ osx()
 	echo "Everything looks good to go!"
 	echo "Compiling for qemu complaings about kvm"
 	echo "You need to run make qemu_no_kvm"
+	rm bootstrap.sh
 }
 
 archLinux()
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -196,7 +196,7 @@ do
 		b) branch="$OPTARG";;
 		e) emulator="$OPTARG";;
 				p) defpackman="$OPTARG";;
-		\?) echo "I don't know what to do with that option..."; exit;;
+		\?) echo "I don't know what to do with that option, try -h for help"; exit;;
 	esac
 done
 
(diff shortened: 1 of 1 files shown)


polymetric1:
I'd say `make qemu_no_kvm` shouldn't just be advised for OSX. KVM hasn't worked for me on Ubuntu or Linux Mint.

polymetric1:
Also, it doesn't really matter (unless you're OCD like me) but i believe it is more of a convention to write commits in present-tense ("do" rather than "did", "fix" rather than "fixed", etc.) and it is suggested in CONTRIBUTING.md.

DomThePorcupine:
@polymetric1 Thank you for the advice on the commit messages. I do not think that they have ever been my strong suit :/ but I will try to be more precise in my wording from now on! Was this most recent one okay?

polymetric1:
I'd maybe say something like `Suggest make qemu_no_kvm for all OSes", but yeah it's fine. Also, I can tell when you commit from your computer as opposed to online on the github website because you haven't set your user info config options. It's probably been asking you to do `git config --global user.name <username>` and `git config --global user.email <email>`.

polymetric1:
I was looking for a shell script formatter and found [this site](http://www.shellcheck.net). Couldn't hurt to run the bootstrap script through. I'm on my phone right now so I can't do it.

polymetric1:
I'd also like to recommend working on a separate branch, so you still have a local master branch that's even with the upstream. It's useful to have around, you may need to `git diff` or test a more stable build. And plus, that's what branches are for.

PR #404, opened 2015-12-30
[WIP] Some fixes to ATA DMA driver

henrikhodne:
**NOTE**: This implementation is broken. I'm mainly opening this to get some feedback if this is worth fixing (I know AHCI things are happening, so not sure if the DMA stuff will continue to live?) and also if anyone can see what bugs there are here. The issue I'm seeing is that the first DMA read request seems to work (the ELF header of the first binary looks correct, but then after that the reads look wrong, although I haven't looked too closely at the exact data read).

This is an attempt to get Redox to boot in VirtualBox using the ATA driver in DMA mode. Some reads that were previously failing are now working, but other ones that previously worked are now failing.

The main change here is that we send an ATA command for each PRD in the PRDT, instead of a single command for the entire section to be read.

There is also a change to separate out the EOT u16 part in the Prd from the size (the size can only be u16 anyways), and using the right bytes for HDDEVSEL (the old bytes were for the 28-bit LBAs instead of the 48-bit LBAs that we use.

### DIFF
--- a/kernel/disk/ide.rs
+++ b/kernel/disk/ide.rs
@@ -169,12 +169,6 @@ const ATA_REG_CONTROL: u16 = 0x0C;
 const ATA_REG_ALTSTATUS: u16 = 0x0C;
 const ATA_REG_DEVADDRESS: u16 = 0x0D;
 
-struct DMARequest {
-    lba: u64,
-    sectors: u16,
-    read: bool,
-}
-
 /// A disk (data storage)
 pub struct Disk {
     base: u16,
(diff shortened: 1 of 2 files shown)


henrikhodne:
With the last commit I'm able to get a shell again, but any commands I've tried stalls and I can't switch to the GUI.
2015-12-30-07-22-33_1066x503

henrikhodne:
I removed the main change, which was sending multiple read requests to the underlying ATA device. After reading more closely I think I misunderstood the original text. This is now booting again on QEMU, and some of the original ELF errors on VirtualBox are gone, though some remain.

(I can rebase all of this at the end)

jackpot51:
I have merged these changes into disk_trait, where I am working on AHCI support in the file scheme

PR #407, opened 2015-12-31
rtl8139: remove magic numbers

henrikhodne:
This removes the many magic numbers for various ports and flags and gives them names, and also switches to using Pio objects directly instead of the inb/outb methods that have a deprecation notice on them
right now (or at least "TODO: Remove this", which is close enough to a deprecation notice).

### DIFF
--- a/kernel/network/rtl8139.rs
+++ b/kernel/network/rtl8139.rs
@@ -11,6 +11,7 @@ use common::{debug, memory};
 
 use drivers::pciconfig::PciConfig;
 use drivers::pio::*;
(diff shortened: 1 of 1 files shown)


henrikhodne:
There shouldn't be any behaviour changes in this. The image boots up just fine, and the PCI info output during boot using qemu is identical, and pings still work when I use net=tap. I haven't found an easy way to test that other network connectivity is working, so I haven't been able to do in-depth testing.

jackpot51:
This is *fantastic* work. Glad you found the net=tap, arping and ping are the best tests so far

PR #413, opened 2016-01-01
Submodulize sodium

Ticki:


polymetric1:
The travis build failed because it was using SSH, it should be using HTTPS (AFAIK)

Ticki:
Fixed now, @polymetric1

PR #426, opened 2016-01-04
[PCI binding] Use match instead of if-else sequense

imp:
This commit completes simple conversion to ```match``` away from ```if-else``` (#419). More serious work will be based on this. Eventually I want it to be a table/map/dictionary driven. I am not sure how to do it yet.


jackpot51:
Good job!

PR #428, opened 2016-01-04
Barfing my work on ZFS

tedsta:
Here it is. Still no write support. I got my work to at least compile by commenting a bunch of stuff out. I will continue working on ZFS, but unfortunately it will have to slow down significantly as I need to shift gears and work on other things in my life (I will now be putting in only ~10 hours per week). I'm about to begin my 8th and final semester of my BS in CS, and so I need to lock down a job.

The road to write support:
- Implement DMU, which handles the in-memory representations of the on-disk data structures and hands reads and writes off to the ARC.
- Finish ZIO and make the ARC use it rather than directly accessing the disk.
- Once DMU and ZIO are done, we can finish the spacemap, metaslab, and vdev systems and finally get some write support.

Cheers


tedsta:
Hmmm something went wrong, my branch is still way behind. I could've sworn I rebased and merged. I'll close this and submit a new one once I figure stuff out.

PR #430, opened 2016-01-04
ZFS Barf

tedsta:
Here it is. Still no write support. I got my work to at least compile by commenting a bunch of stuff out. I will continue working on ZFS, but unfortunately it will have to slow down significantly as I need to shift gears and work on other things in my life (I will now be putting in only ~10 hours per week). I'm about to begin my 8th and final semester of my BS in CS, and so I need to lock down a job.

The road to write support:

- Implement DMU, which handles the in-memory representations of the on-disk data structures and hands reads and writes off to the ARC.
- Finish ZIO and make the ARC use it rather than directly accessing the disk.
- Once DMU and ZIO are done, we can finish the spacemap, metaslab, and vdev systems and finally get some write support.

Cheers


Ticki:
Cool work :+1:

filesystem/apps/zfs/vdev_label.rs
```diff
@@ -0,0 +1,1011 @@
...
+
+// Size of label regions at the start and end of each leaf device.
+const VDEV_LABEL_START_SIZE: usize = (2 * mem::size_of::<VdevLabel>() + VDEV_BOOT_SIZE);
+const VDEV_LABEL_END_SIZE: usize = (2 * mem::size_of::<VdevLabel>());
+const VDEV_LABELS: u8 = 4;
+const VDEV_BEST_LABEL: u8 = VDEV_LABELS;
+
+// Basic routines to read and write from a vdev label.
+// Used throughout the rest of this file.
+vdev_label_offset(psize: u64, l: u8, offset: u64) -> u64 {
```
Ticki: Why are you writing C/Rust mix here?

tedsta: Whoops, that was not supposed to get pushed. That file is no longer necessary. There might be other unnecessary files too. I'll submit another PR to remove them.

PR #431, opened 2016-01-05
IRQ performance improvement, add CSO exception, more ANSI support

Ticki:


kernel/main.rs
```diff
@@ -131,29 +131,23 @@ static PIT_DURATION: Duration = Duration {
...
-                }
+        for i in env().contexts.lock().iter() {
+            if i.interrupted {
+                halt = false;
+                break;
             }
         }
 
+        asm!("sti" : : : : "intel", "volatile");
+
```
jackpot51: To make this not halt after an interrupt, hlt needs to be the instruction immediately after sti

Ticki: Woosh. Will fix immediately.

kernel/main.rs
```diff
@@ -131,24 +131,20 @@ static PIT_DURATION: Duration = Duration {
...
-        {
-            let contexts = ::env().contexts.lock();
-            for i in 1..contexts.len() {
-                if let Some(context) = contexts.get(i) {
-                    if context.interrupted {
-                        halt = false;
-                        break;
-                    }
-                }
+        for i in env().contexts.lock().iter() {
```
jackpot51: Context 0 is this function, use .skip(1) on the iter or you will not halt

Ticki: Really. I thought it wouldn't iterate over context 0.

PR #434, opened 2016-01-05
Hot/cold precedence optimization on interrupt handling

Ticki:


jackpot51:
0x20 is the hottest because it is the programmable interrupt timer

jackpot51:
Use `cat interrupt:` to see the interrupt counts

jackpot51:
I did pull the master sodium though

Ticki:
thanks

PR #451, opened 2016-01-11
Fix failure of bootstrap script as reported by @sfisol

Ticki:


jackpot51:
Is this really necessary? It seems like a very messy hack

Ticki:
Apperently, it was a problem with @sfisol's config

jackpot51:
Ok, thanks

PR #452, opened 2016-01-11
Add note on bandwidth

Ticki:


polymetric1:
What does the **NB:** mean?

Ticki:
https://en.wikipedia.org/wiki/Nota_bene

PR #456, opened 2016-01-14
Rewrite Pio{8,16,32} as generics

imp:
While working on PCI stuff I noticed that while MMIO is implemented as generics, PIO isn't. Since I am trying to use generics in PCI code as well I thought it is a good opportunity to convert the PIO to use generics as well. [*Little did I knew about how to do it right. However, I am glad I went through it as I learned a lot. Special thanks to @Ticki for helping me out!*]

Anyway,
this work is splitted into 5 commits. Each does only one thing for the ease of review. I do appreciate review by the way.

I believe the new interface in pio.rs is a little bit cleaner. And it resembles mmio.rs now, which is nice. I think I will be able to unify it even more eventually.

Also as a nice side effect I was able to drop some number of ```unsafe``` blocks.

### DIFF
--- a/kernel/audio/ac97.rs
+++ b/kernel/audio/ac97.rs
@@ -44,18 +44,18 @@ impl Resource for AC97Resource {
         unsafe {
             let audio = self.audio as u16;
 
-            let mut master_volume = Pio16::new(audio + 2);
-            let mut pcm_volume = Pio16::new(audio + 0x18);
+            let mut master_volume = Pio::<u16>::new(audio + 2);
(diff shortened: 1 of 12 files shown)


imp:
I am not sure what happens with CI checks - I have it working on my OSX

polymetric1:
I believe Travis builds on Linux and AppVeyor is on Windows, so it must only be working on OSX

imp:
Travis builds on both Linux and OSX. According to logs test passed on Linux. Stalled for 10m on OSX and was killed and completely went nuts (didn't find the linker) on Windows.

PR #459, opened 2016-01-15
Update README.md

llambda:
Add step to update git submodules


polymetric1:
Thanks, can't believe I didn't notice this

llambda:
np! BTW, I just pushed an updated commit, I was missing the $ at the beginning of the line.

PR #462, opened 2016-01-16
Getting started with implementing unit testing for Redox

stratact:
**Work-In-Progress/Do Not Merge**

This is an attempt to add support for unit testing in Redox by adding `make test` which will allow the user to run a series of `#[test]`s to verify functionality and possibility of bugs.

Currently I have only the unit testing for `GetSlice` but more files with unit testing can be easily added

Here are the following that are preventing this PR from being mergeable, is the need of:
- [x] `std::io::prelude::*` (libredox)
- [x] `std::sync::Once` (libredox)
- [ ] `std::sync::StaticMutex` (libredox)
- [ ] Wrap-up cleaning the Makefile

And possible other dependencies of dependencies to make the `libtest` crate happy.

More work still needs to be done for this to be useful.


Ticki:
Neat. What about quickchecking too?

jackpot51:
I want to get this to compile, so I will merge, as orbital branch can break

Ticki:
@jackpot51 It compiles.

Ticki:
1 second apart...

jackpot51:
Try `make test` ticki

Ticki:
That's odd. Seem to be an issue with the makefile.

jackpot51:
I don't think test code will work inside the kernel, but it doesn't work in libraries either right now, so this work will eventually fix it

Ticki:
Why shouldn't it work in the kernel?

jackpot51:
We'll see, some will, some code like driver tests must not use #[test] but be called when the kernel is booted

Ticki:
Yeah, that's right. A work-around is to have a function only called when `cfg!(test)`, which is executed right after the system is booted.

stratact:
Guys... there was a reason why I had a `Wrap-up cleaning the Makefile` check box, the Makefile was in testing stages before `make test` would be useable. I was testing `make $(BUILD)/libtest.rlib` to make sure libtest dependencies were happy.

stratact:
I'll make another PR if there still needs cleaning up.

PR #467, opened 2016-01-16
StaticMutex for you, stratact

Ticki:
cc @stratact


stratact:
Thanks, much appreciated.

PR #468, opened 2016-01-16
Reduce libtest errors

stratact:
I have also added libtest as the dependency to `make test` to make it useable but it still errors due to how incomplete our libredox/libstd implementation is.


Ticki:
:+1:

PR #470, opened 2016-01-19
Simplify the GetSlice implementation

Ticki:
Entails:

- Let it take a type implementing `RangeArguments` (you can use the range syntax, `a..b`, `a..`, `..b` and `..`).
- Slightly faster by using branches and comparations
- A bit simpler code

cc @stratact

### DIFF
--- a/kernel/common/parse_ip.rs
+++ b/kernel/common/parse_ip.rs
@@ -19,10 +19,14 @@ pub fn parse_port(string: &str) -> &str {
         }
     }
 
-    string.get_slice(string.find(':').map(|a| a + 1), Some(b))
+    if let Some(a) = string.find(':').map(|a| a + 1) {
+        string.get_slice(a..b)
+    } else {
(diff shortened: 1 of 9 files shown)


jackpot51:
I like using the range arguments a lot

kernel/common/parse_ip.rs
```diff
@@ -19,10 +19,14 @@ pub fn parse_port(string: &str) -> &str {
...
+        string.get_slice(a..b)
+    } else {
+        string.get_slice(..b)
+    }
 }
 
 /// Get the host from a string (ip)
 pub fn parse_host(string: &str) -> &str {
-    string.get_slice(None, string.find(|c| c == ':' || c == '/').map(|b| b + 1))
+    string.get_slice(..string.find(|c| c == ':' || c == '/').map(|b| b + 1).unwrap_or(0))
```
jackpot51: This should be unwrap_or(string.len())

PR #474, opened 2016-01-19
Reapply the old `GetSlice` as `OptionSlice`, because sometimes it's needed to use `Option<usize>` parameters

stratact:
This is the return of the old `GetSlice`, now known as `OptionSlice`, because when times your just returning an `Option<usize>`, say from using `.get()` or `.find()` method slices, you should have the work done for you and done correctly without you having to think about it. So here are the following changes:

* Adds `OptionSlice` which takes `Option<usize>` as parameters
* Moves both `GetSlice` and `OptionSlice` to a `slice` module
 * And publicly exports them there
 * Fix the imports for other files
* Use `OptionSlice` in the more appropriate places where `GetSlice` is not well suited.


stratact:
@jackpot51 looks like we need to update the rust submodule?

stratact:
Closing for now. I want to consult with @Ticki more about a possible better solution.

PR #475, opened 2016-01-20
OptionSlice redux (yes, I spelled the 2nd word correctly :P)

stratact:
For the same reasons in PR #474 and after much thought, I was able to make `.option_slice()` take a range much like `get_slice()` except it takes `Option<usize>` for both start and end of the range: So here are the following changes:

* Adds `OptionSlice` which takes `Option<usize>` as **range** parameters
* Moves both `GetSlice` and `OptionSlice` to a `slice` module
* Use `OptionSlice` in the more appropriate places where `GetSlice` is not well suited
* Cleanup unused imports in `GetSlice`


jackpot51:
Cool, I like having both so you can pass find straight to get_slice

stratact:
Alright, will attempt that at the next PR.

jackpot51:
Is it possible?

jackpot51:
To combine both into one?

stratact:
@LazyOxen actually attempt at it in an spoof version and I cleaned it up afterwords. I think I would be able to adapt it for `GetSlice`.

jackpot51:
cool

Ticki:
:+1:

PR #476, opened 2016-01-20
GetSlice can now use both `usize` and `Option<usize>`

stratact:
* Along with removing the now deprecated `OptionSlice`
* Implemented the `AsOption` trait for converting either `usize` or `Option<usize>` to `Option<usize>` (credit goes to @LazyOxen for writing out the functionality in the Rust Playpen)

So you can have your :cake: and eat it too.

The only downsides to doing this are:
* That you cannot do a `usize..Option<usize>` or `Option<usize>..usize` because it's statically dispatched
 * So it must be either `usize..usize` or `Option<usize>..Option<usize>`
* The `test!(array.get_slice(..) == &array);` test errors, because when using `..` ranges, `GetSlice` needs to know the type parameters explicitly and I didn't feel like with filling it in for now.

But other than that, [I Feel Good!](https://www.youtube.com/watch?v=B1wOK9yGUYM)


Ticki:
:+1:, nice.

stratact:
Thinking about it again, you wouldn't need to do `array.get_slice(..)` you could just do `&array[..]` which the same as `&array` anyway :stuck_out_tongue_winking_eye:

PR #488, opened 2016-01-31
RUSThello to Redox

EGhiorzi:
First commit to port RUSThello to Redox.


Ticki:
Freakin' awesome work!

PR #491, opened 2016-02-03
Avoid dynamic dispatch, allocations, clones, and other expensive operations on common objects and structures in the kernel (Scheme, KScheme, URL, Resource, etc.)

Ticki:
WIP. Do not merge.

There are still some bugs (and errors?). If anyone wants to play around with it, feel free to do so.


Ticki:
So .. many .. allocations. I'm tired now. Good night.

jackpot51:
Merged this into no_kernel_alloc branch

jackpot51:
Working on making it run correctly now

jackpot51:
A warning: Some things must, must, must be allocated. If I find things like this I will make a note of them here

jackpot51:
Honestly, I prefer to do this a different way

Ticki:
@jackpot51 Could you expand on what you think is wrong?

jackpot51:
I believe it to be impossible to have context have a lifetime, I believe it must be fully owned. We can certainly replace Url with &str up to that point, but it must be allocated to pass context memory boundaries, like in the execute function.

Ticki:
@jackpot51 I get your point. I'll remove the context lifetime.

Ticki:
@jackpot51 I updated it know, however it is still incomplete. When I get time, later, I'll try to fix it.

jackpot51:
That's good. The name can't be static str because it comes from a non-static argument provided in the execve syscall.

jackpot51:
This is still available in the kernel_no_alloc branch. I am closing because I am working out a way to minimize kernel allocations along with page table fixes in another branch

Ticki:
Sure.

PR #493, opened 2016-02-04
improved assembly slightly

Roxxik:


Roxxik:
Ops, this doesn't work currently, wrongly tested, do not merge!

Roxxik:
Fixed it, works just as before, but might be slightly faster, because rep is microcoded (on real hardware)

jackpot51:
Cool, keep looking for more of these!

PR #500, opened 2016-02-06
Getting unreal

Roxxik:
- [x] i386
- [ ] x86_64


Roxxik:
i386 works(it boots) but i haven't tested much of it
x86_64 loads the kernel and jumps into it, but something is still wrong

Roxxik:
fixed the history.
currently it crashes after starting the HID driver

jackpot51:
Very impressive work. Since x86_64 was not working anyways, I will merge

PR #505, opened 2016-02-10
get_dir refactoring

k0pernicus:


libstd/src/env.rs
```diff
@@ -76,6 +75,17 @@ pub fn current_dir() -> Result<PathBuf> {
     }
 }
 
+/// Method to return the current directory
+pub fn current_dir() -> Result<PathBuf> {
+    // Return the current path
+    get_dir("./")
```
jackpot51: Should this be get_path_from ?

k0pernicus:
Right! Sorry for that mistake...

PR #508, opened 2016-02-11
entering kernel correctly, need to change paging code

Roxxik:
Do not merge yet, the kernel paging code has to be updated to PAE first


jackpot51:
Merged into https://github.com/redox-os/redox/tree/higher-half, make any other PRs there

PR #519, opened 2016-02-24
Refactor the schemes more, especially with path() methods

stratact:


kernel/schemes/file.rs
```diff
@@ -41,22 +41,18 @@ impl Resource for FileResource {
         })
     }
 
-    fn path(&self, buf: &mut [u8]) -> Result <usize> {
-        let mut i = 0;
-
+    fn path(&self, buf: &mut [u8]) -> Result<usize> {
         let path_a = b"file:/";
```
jackpot51: You could get an iter_mut() at the top of this path function, then use it without having to skip

PR #522, opened 2016-02-28
no_alloc take Ⅲ

Ticki:
Now, there are three URL representations, all of which tracks both the scheme and the reference, such that reparsing is unnecessary:

1. `Url`: A struct with two string slices. It is obviously unowned, and you have to track the lifetime. You can convert a `&str` to an `Url` in `O(n)`, it simply requires finding the `:`, and taking two slices. This can be casted into the forms below by using one of the built-in methods.

2. `OwnedUrl`: This keeps two heap-allocated, owned `String`s representing the scheme and the reference. This is useful in a number of scenarios (when `Url` isn't sufficient). There is currently relatively few methods defined on it, so to manipulate it, use `as_url()`, which converts the `OwnedUrl` to an `Url`.

3. `CowUrl`: An Copy-On-Write (CoW) URL, which is useful whenever heap allocation is conditional, that is, heap allocation is only necessary in a limited number of cases.

`Url` is used for arguments. `OwnedUrl` is used for unconditional manipulation. `CowUrl` is used for conditional manipulation.


Ticki:
Wow, sumodules are annoying.

Ticki:
Damn, it does not build on the latest nightly rust:

```
libstd/src/lib.rs:178:9: 178:26 error: extern crate `core_rand` is private, and cannot be reexported (error E0364), consider declaring with `pub`, #[deny(private_in_public)] on by default
warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
```

Ticki:
Note that this PR is not introducing errors, these errors are already in master (if you build with the latest nightly).

jackpot51:
Thanks Ticki!

Issue #525, opened 2016-02-28
Documentation, transparency, and information


Ticki:
[x] Contribute.md
[ ] Complete design document.
[ ] Formal proof of the design (I'm working on this)
[ ] `#![deny(missing_doc)]` in all Redox-related crates.
[ ] A overview of the design principles.
[ ] "Everything is an URL" explanation (I'm working on this)
[ ] "What is done, and what is not done"
[ ] Complete list of goals and non-goals.
[ ] What is the rationale behind using Rust? What advantages does it carry? What disadvantages?
[ ] How and where is the userspace/kernelspace seperation enforced?

TODO: Add more

Ticki:
https://github.com/redox-os/book

Ticki:
Just completed some of them.

Ticki:
I checked some of them off.

Ticki:
See #610.

PR #534, opened 2016-03-03
docgen - a manpage generator

Ticki:
This is a very simple manpage generator. Manpages are extracted from the codebase, and started by `@MANSTART{name}` and ended by `@MANEND`. These are then placed in filesystem/man, which can then later be read using `cat`.


Ticki:
A part of the documentation effort, #525.

PR #541, opened 2016-03-05
Use 'to_owned()' instead of 'to_string()'

Ticki:


Ticki:
Note that it builds, but, for some reason, Travis failed in the installation process.

PR #546, opened 2016-03-06
RUSThello update to v1.1.0

EGhiorzi:
Lots and lots of changes everywhere


EGhiorzi:
"All checks have failed" doesn't look good… I have no clue what the issue is.

filesystem/apps/rusthello/src/interface.rs
```diff
@@ -0,0 +1,299 @@
...
+				println!("\nGoodbye!\n\n\n");
+				return UserCommand::Quit;
+			}
+			_ => println!("This is not a valid command!"),
+		}
+	}
+}
+
+pub fn new_player(side: reversi::Disk) -> Option<players::Player> {
+	loop {
```
Ticki: You could use a for-loop here (by letting get_user_command be an iterator instead).

Ticki:
Noice!

filesystem/apps/rusthello/src/interface.rs
EGhiorzi: Sure, I'll need to give a look at iterators

EGhiorzi: Sorry @Ticki I think I totally misunderstood your idea. I can implement an iterator like this
```
impl Iterator for UserCommand {
	type Item = UserCommand;

	fn next(&mut self) -> Option<UserCommand> {
            \\ returns a UserCommand matching user's input
        }
    }
}
```
but it seems wrong to me: it's not `UserCommand` which should produce a next `UserCommand` (and indeed `&mut self` is totally useless). Moreover I can't see a natural way to write the for-loop with this iterator.

Ticki: `UserCommands` could hold a handle to `stdin`, which could be read from once `next()` is called. This is faster, but perhaps not worth it?

EGhiorzi: Performance-wise it should make no difference, that's not a critical point in the code. I trust you on that being a better solution, but I'm not sure I know how to do it… Maybe I'll look at it again when I learn a bit more of Rust. If you'd like to fix it yourself though feel free to do it, of course!

Ticki: Well, reconstructing and locking stdin on every user input can be quite inefficient. It makes no big difference in this case, though.

I'll try to look into Rusthello when I get time :)

EGhiorzi: Oh I think I finally got what you meant, I'll try that!

jackpot51:
I haven't merged this yet because it breaks the build, The makefile needs to be updated to point to the right main.rs

jackpot51:
I am merging now

EGhiorzi:
I always make a mess with git. If you can fix it without effort, that would be grat. Otherwise I'll try to fix it tonight and resubmit the pull request.

PR #570, opened 2016-03-21
Don't try to load non-evisting redoxfs.bin

genodeftest:
Currently redoxfs.bin isn't built for `make all`, thus trying to load /etc/redoxfs.bin at runtime will always fail.

Don't load redoxfs.bin

remove loading redoxfs.bin

**Drawbacks**: [if any, describe the drawbacks of this pull request.]
If someone is testing redoxfs.bin, he might want to add this line again.


jackpot51:
Thanks

Issue #572, opened 2016-03-22
Tracking issue for libstd/libredox compatibility.


Ticki:
**Description**: This issue tracks the compatibility between Redox's libstd ("libredox") and the upstream libstd.

**List**:

- [x] Full IO compatibility. [Done](https://github.com/redox-os/redox/pull/552).
- [x] Moving extras to a seperate library. [Done](https://github.com/redox-os/extras).
- [ ] Full fs compatibility.
- [ ] Lacking sync primitives:
  - [x] Mutex.
  - [x] mpsc.
  - [ ] Barrier.
  - [ ] CondVar.
  - [x] Atomic types.
- [ ] Remove unnecessary reexports.
- [x] Collections:
  - [x] HashMap.
    - [x] SipHasher.
  - [x] HashSet
  - [x] BTreeMap.
  - [x] BTreeSet.
  - [x] Vec.
  - [x] VecDeque.
  - [x] LinkedList.
- [ ] Networking:
  - [x] CheckSum.
  - [x] IPv4.
  - [x] Ipv6.
  - [ ] Incoming.
  - [ ] TcpListener.
  - [ ] TcpStream.
  - [ ] UdpSocket.
  - [ ] LookupHost
- [ ] File management:
  - [x] File.
  - [x] canonicalize.
  - [x] create_dir.
  - [x] metadata.
  - [x] read_dir.
  - [x] remove_dir.
  - [x] remove_file.
  - [x] OpenOptions.
  - [ ] copy.
  - [ ] rename.
  - [ ] set_permissions.
  - [ ] soft_link.
  - [ ] hard_link.
  - [ ] read_link.
  - [ ] create_dir_all.
  - [ ] remove_dir_all.
  - [ ] Permissions.
  - [ ] DirBuilder.
- [ ] Time:
  - [x] Duration.
  - [x] Instant.
  - [x] SystemTime.
  - [ ] SytemTimeError.
- [ ] Threading:
  - [x] spawn.
  - [x] sleep.
  - [x] yield_now.
  - [x] .join().
  - [ ] park.
  - [ ] park_timeout.
  - [ ] panicking.
  - [ ] LocalKey.
  - [ ] Builder.
  - [ ] current.

Ticki:
cc @Mouvedia.

Mouvedia:
redox-os/extras doesn't exist. Should it be [redox-os/libextra](https://github.com/redox-os/libextra) instead?

nounoursheureux:
[...]

PR #574, opened 2016-03-22
Add fs::copy and fs::rename

nounoursheureux:
I added the `copy` and `rename` functions to the `fs` module. See #572

### DIFF
--- a/libstd/src/fs.rs
+++ b/libstd/src/fs.rs
@@ -1,6 +1,6 @@
 use core::ops::Deref;
 use core_collections::borrow::ToOwned;
-use io::{Read, Error, Result, Write, Seek, SeekFrom};
+use io::{self, Read, Error, Result, Write, Seek, SeekFrom};
 use os::unix::io::{AsRawFd, FromRawFd, IntoRawFd, RawFd};
 use mem;
 use path::{PathBuf, Path};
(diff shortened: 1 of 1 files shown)


libstd/src/fs.rs
```diff
@@ -343,6 +343,17 @@ pub fn create_dir<P: AsRef<Path>>(path: P) -> Result<()> {
     }
 }
 
+pub fn copy<P: AsRef<Path>, Q: AsRef<Path>>(from: P, to: Q) -> Result<u64> {
+    let mut infile = try!(File::open(from));
+    let mut outfile = try!(File::create(to));
+    io::copy(&mut infile, &mut outfile)
+}
+
+pub fn rename<P: AsRef<Path>, Q: AsRef<Path>>(from: P, to: Q) -> Result<()> {
```
Ticki: This is not how rename should work.

jackpot51: True, we need a sys_rename, but it is ok for now

jackpot51:
I will start working on sys_rename

jackpot51:
Somebody should update mv to use fs::rename

Ticki:
> Somebody should update mv to use fs::rename

Done

PR #602, opened 2016-04-01
Add environment variables

nounoursheureux:
**Problem**: Redox doesn't have environment variables. Each process should have its own set of environment variables, that are copied to the child when the process forks.

- Add the `env_vars` field to the  `Context` struct
- Add the `env` scheme to access environment variables from userspace
- Add the `env::var`, `env::set_var`, `env::remove_var` and `env::vars` functions to `libstd`

**TODOs**:
- ~~Add a way to remove environment variables (I'm working on it)~~
- Maybe add the `export` command to extrautils.
- Set the `HOME`, `PWD`, etc. variables on login

**State**: Ready

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -197,8 +197,7 @@ extrautils: \
 	filesystem/bin/grep \
 	filesystem/bin/less \
 	filesystem/bin/mtxt \
-	filesystem/bin/rem \
-	filesystem/bin/envtest
+	filesystem/bin/rem
 
 filesystem/bin/%: crates/games/src/%/main.rs crates/games/src/%/*.rs $(BUILD)/libextra.rlib $(BUILD)/libtermion.rlib
 	mkdir -p filesystem/bin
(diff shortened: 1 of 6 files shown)


nounoursheureux:
I added the `env::remove_var` function

nounoursheureux:
This should fix the build

jackpot51:
Given the way this requires syscalls for every variable operation, I don't see it as the final design, but it is definitely better than nothing

Thanks!

nounoursheureux:
Thanks for merging ! I'm pretty new to OS dev, so could you explain how the userspace and the kernel could communicate without syscalls ?

jackpot51:
In Linux, for example, the args are passed in the execve syscall
The process copies the args into allocated memory so that it can modify them
A new process that is launched can be passed the current args, or modified args, using the execve syscall

Ticki:
@jackpot51 It should be noted that that model have numerous downsides.

jackpot51:
@ticki Then you should note the pros/cons of each type

jackpot51:
Well, thanks to this, the terminal size can now be retrieved using env, which allows it to be retrieved for the orbital terminal

PR #608, opened 2016-04-08
Introduce a new Redox-specific syscall, SYS_SUPERVISE.

ticki:
**Problem**: You cannot gain fine-grained control over a child process in Redox. This makes it possible to do things which are clasically covered by ptrace, seccomp, chroot and friends: to run another process in a restricted, traced, and supervised environment, which is useful for various purposes, such as emulation, virtualisation, logging, and debugging.

**Solution**: Introduce a new system call, SUPERVISE, which allows the parent process to handle the system calls of the child process, by blocking the child process when system calls are received.

- Introduce a system call, SUPERVISE:

 SUPERVISE takes a PID specifing the process to be supervised. This PID must be a child process of the invoker. If not, EACCES will be returned.

 The process identified by the given PID will be restricted in such a way, that every syscall
made will mark the process as blocked and store the syscall until it is handled by the parrent.

 The return value (if successful) is a file descriptor, from which syscalls can be read and written: the syscalls are read in `Packet` sized packages, containing the respective blocking syscall. If no syscall is blocking (or the last blocking syscall have been handled), 0 bytes will be read to the buffer.

 Writing pointer sized integers to this file handle will set the EAX register of the particular
process, after which the process is unblocked and the syscall buffer is emptied. The behavior of writing packages of unexpected size is unspecified.

 Note that a process blocked by a syscall will have its potential sleep cleared (i.e., it will not wake up after the sleep is finished).

 Passing a non-existent PID results in ESRCH. 

- Add a flag, _CLONE_SUPERVISE_, to SYSCLONE. This flag will make the clone enter supervised mode immediately.
 This means that the process can run in supervised mode, even not being connected to a supervisor yet. In other words, the parent can later on supervise the process and handle the potential blocking syscall.

 This is an important security measure, since otherwise the process would be able to fork it self right after starting, making supervising it impossible.

- Add a `spawn_supervised` method to `Command`, which will spawn the command with the CLONE_SUPERVISED flag.

- Slightly refactor the `Command` internals to allow specifing arbitrary flags to SYS_CLONE.

**Drawbacks**: Increased complexity and surface area. One might argue that it, on a long-term basis, does the opposite: namely allow us to retain compatibility without introducing extra syscalls into the kernel.

**TODOs**:

- Security analysis.
- Add a `strace` tool to extrautils.
- Add a restrict tool to extrautils.
- Create a Linux compatibility layer.
- Allow handling certain non-0x80 interrupts?

There are no _known_ issues otherwise.

**State**: Ready, though it require future work (see TODO).

**Other**: sry for the eastereggs

### DIFF
--- a/crates/system/scheme.rs
+++ b/crates/system/scheme.rs
@@ -1,4 +1,3 @@
-
 use core::ops::{Deref, DerefMut};
 use core::{mem, slice};
 
--- a/crates/system/syscall/redox.rs
+++ b/crates/system/syscall/redox.rs
@@ -1,8 +1,40 @@
-use syscall::arch::syscall2;
+use syscall::arch::{syscall2, syscall1};
 use error::Result;
 
 pub const SYS_DEBUG: usize = 0;
+pub const SYS_SUPERVISE: usize = 1638; // loominatzi confirmed
 
 pub fn sys_debug(buf: &[u8]) -> Result<usize> {
     unsafe { syscall2(SYS_DEBUG, buf.as_ptr() as usize, buf.len()) }
+}
+
+/// <!-- @MANSTART{supervise} -->
+/// Supervise a given child process' system calls.
+///
+/// SUPERVISE allows a process to run another process in a restricted, traced, and supervised
+/// environment, which is useful for various purposes, such as emulation, virtualisation, tracing,
+/// logging, and debugging.
+///
+/// SUPERVISE takes a PID specifing the process to be supervised. This PID must be a child process
+/// of the invoker. If not, EACCES will be returned.
+///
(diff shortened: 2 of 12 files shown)


jackpot51:
Its going to take me some time to get around to this, just to warn you

Ticki:
Sure, I don't mind.

jackpot51:
It looks very good! I lefy a couple notes

jackpot51:
I may merge and write a test tool before you come back

Ticki:
FYI, the many of the patches you did in the `supervise` branch are wrong and can be bypassed.

Ticki:
I pushed some fixes to address your comments.

Ticki:
I fixed the bugs.

PR #611, opened 2016-04-10
Add Gentoo Linux to bootstrap.sh

dalance:
**Problem**:
`bootstrap.sh` don't work on Gentoo Linux.

**Solution**:
Add `gentoo()` function for installing required packages, and call `gentoo()` function when `emerge` command exits. (`emerge` is package manager of Gentoo Linux.)

**Drawbacks**:
QEMU installation is not automated.

**State**:
Ready

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -171,6 +171,32 @@ suse()
 	sh redox/setup/binary.sh
 }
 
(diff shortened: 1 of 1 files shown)


jackpot51:
Thanks!

Issue #620, opened 2016-04-22
Build Fails (Manjaro)


BunnyApocalypse:
**Reproduction**: [describe how you are able to reproduce ("trigger") this bug/issue.]

I attempted to install rust using the auto install script. 

**Expected behavior**: [describe the behavior you would expect the repro to yield.]

It installs properly 

**Actual behavior**: [describe the actual behavior, which is presented through the repro.].

At the end of the install script it returns this before telling me to make make all etc it gives me this error

`
Travis reports that the last build *FAILED* :(
Might want to check out the issues before building
`
Then, when i do cd into the redox directory and make all it returns 
```
mkdir -p build/i386-unknown-redox/debug
RUST_BACKTRACE=1 rustc --target=i386-unknown-redox.json -L build/i386-unknown-redox/debug -C no-prepopulate-passes -C no-stack-check -C opt-level=2 -Z no-landing-pads -A dead_code -o build/i386-unknown-redox/debug/libcore.rlib rust/src/libcore/lib.rs
error: internal compiler error: unexpected panic
note: the compiler unexpectedly panicked. this is a bug.
note: we would appreciate a bug report: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#bug-reports
note: run with RUST_BACKTRACE=1 for a backtrace
thread 'rustc' panicked at 'Field data-layout in target specification is required', ../src/librustc_back/target/mod.rs:370
stack backtrace:
   1:     0x7f162121a9f0 - std::sys::backtrace::tracing::imp::write::h9fb600083204ae7f
   2:     0x7f162122811b - std::panicking::default_hook::_$u7b$$u7b$closure$u7d$$u7d$::hca543c34f11229ac
   3:     0x7f1621227cbc - std::panicking::default_hook::hc2c969e7453d080c
   4:     0x7f16211eccaf - std::sys_common::unwind::begin_unwind_inner::h30e12d15ce2b2e25
   5:     0x7f16211eed98 - std::sys_common::unwind::begin_unwind_fmt::hb2de8a9968d38523
   6:     0x7f161d2800ca - rustc_back::target::Target::from_json::h0865f3d0efe9d5a8
   7:     0x7f161d2822bf - rustc_back::target::Target::search::load_file::h10b754eb1c29e7bb
   8:     0x7f161d2816ba - rustc_back::target::Target::search::h5d24d442a2ed9dcf
[...]

PR #622, opened 2016-04-23
Add data-layout fields to build targets.

NilSet:
**Problem**: Breaking change in rust-lang/rust#32939 makes the current nightly refuse to build without a data-layout field in the target json. 

**Solution**: [Take the data-layout emitted by clang for the target-triple](https://github.com/rust-lang/rust/issues/31367#issuecomment-213595571) and add it to the target json.

**TODOs**: Validate that clang's defaults are what we want. [Documentation on data-layout format](http://llvm.org/docs/LangRef.html#data-layout)

**Fixes**: #620

### DIFF
--- a/i386-unknown-redox.json
+++ b/i386-unknown-redox.json
@@ -1,7 +1,8 @@
 {
-    "llvm-target": "i686-unknown-redox",
+    "llvm-target": "i386-unknown-redox",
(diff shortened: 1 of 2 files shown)


jackpot51:
Thanks for taking care of this!

PR #625, opened 2016-04-25
Add pkg-config dependency for osx

DomThePorcupine:
pkg-config is needed for building on osx


jackpot51:
Thanks, very quick response!

PR #627, opened 2016-04-26
Implement kernel logs and add the klog scheme

nounoursheureux:
**Problem**: The kernel doesn't keep logs, so if it wants to communicate with the user the only way is to print something on the screen. While this works well for critical errors (because you *want* the user to see them), it is excessive for warnings or simple informations.

**Solution**: Have the kernel keep tracks of info/warning/error logs, and let the users access them when they want to.

- Add the `klog` function which add a message to the kernel logs.
- Add the `LogLevel` enum.
- Add the `klog` scheme

To read the kernel logs you can simply do:
```sh
$ cat klog:
```

**TODOs**:
- Maybe use a circular buffer to avoid using too much memory.
- Add a way read logs only of a certain priority level. Here is an example of the syntax that I'm thinking of:
```bash
$ cat klog:critical/error/warning
```
to view only critical/error/warning logs and
```bash
$ cat klog:!debug/!info
```
What do you think ?
to view everything except debug and info logs.

**State**: Ready.

### DIFF
--- a/kernel/schemes/klog.rs
+++ b/kernel/schemes/klog.rs
@@ -1,0 +1,95 @@
+use fs::{KScheme, Resource, Url};
+use fs::resource::ResourceSeek;
+use collections::string::String;
+use alloc::boxed::Box;
+use system::error::Result;
+use logging::LogLevel;
+
+/// The kernel log scheme.
+pub struct KlogScheme;
+
+impl KScheme for KlogScheme {
+    /// Returns the name of the scheme: "klog"
+    fn scheme(&self) -> &str {
+        "klog"
+    }
+
+    /// Returns a resource. The `url` and `flags` arguments are currently unused.
+    fn open(&mut self, _: Url, _: usize) -> Result<Box<Resource>> {
+        Ok(Box::new(KlogResource {
+            pos: 0,
+        }))
+    }
+
+    /// Clears the logs.
+    fn unlink(&mut self, _: Url) -> Result<()> {
+        let mut logs = ::env().logs.lock();
+        logs.clear();
+        Ok(())
+    }
+}
+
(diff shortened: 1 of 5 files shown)


bjorn3:
Looks like glob is broken

bjorn3:
I have filled an issue: kevinmehall/rust-peg#131

kernel/schemes/klog.rs
```diff
@@ -0,0 +1,95 @@
...
+    }
+}
+
+/// The kernel log resource.
+pub struct KlogResource {
+    pos: usize,
+}
+
+impl KlogResource {
+    fn get_log_str(&self) -> String {
```
jackpot51: Calling this on every read is bound to be slow.

kernel/schemes/klog.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Fills `buf` with the kernel log. Each message is prefixed by its log level:
+    /// - `CRIT`
+    /// - `ERROR`
+    /// - `WARN`
+    /// - `INFO`
+    /// - `DEBUG`
+    fn read(&mut self, buf: &mut [u8]) -> Result<usize> {
+        let mut i = 0;
+        let logs = self.get_log_str();
+        while i < buf.len() && self.pos < logs.bytes().count() {
```
jackpot51: If you are doing byte indexes you can use .len() and [self.pos..]

jackpot51:
I am going to merge so I can add messages. I would consider naming it syslog and allowing other applications to write to it. There must be a memory limit or else there could be issues, I would like to log disk access for example

ticki:
@jackpot51 It should definitely be truncated to avoid endless logs.

PR #630, opened 2016-05-01
Add libfuse to list of required dependencies

nickodell:
**Problem**: I didn't know that libfuse was needed, so I got build errors

**Solution**: Adds documentation to say that you need libfuse-dev


jackpot51:
thanks!

Issue #644, opened 2016-05-15
File browser crashes when going up a directory


ca1ek:
**Reproduction**: Open up the file browser, then attempt to go up a directory by clicking `../`

**Expected behavior**: Showing the parent directory

**Actual behavior**: File browser just closes

**Build information**: rustc 1.10.0-nightly (4ec5ce5e4 2016-05-12),  
qemu-i386 version 2.0.0 (Debian 2.0.0+dfsg-2ubuntu1.22)

**Blocking/related**: n/a

ca1ek:
Fixed by #648

PR #645, opened 2016-05-16
Get file_manager working via ugly hack

NilSet:
**Problem**: file_manager app can't go up into parent directories

**Solution**: trim the leading `file:` from the path returned by `get_parent_directory`

**Drawbacks**: Doesn't address the fact that something somewhere isn't handling paths with the scheme on them correctly.

**TODOs**: Find out where the bug really is.

**Fixes**: #644 

**State**: Ready


jackpot51:
Merged temporarily, I hope we figure out what the issue is

PR #648, opened 2016-05-18
Fix orbital scheme's handling of embedded ':' char

NilSet:
**Problem**: the `orbital:` scheme was incorrectly parsing urls that contained an embedded `:`. This made the file_manager app crash when it tried to set the window title to a path with the `file:` scheme included.

**Solution**: Split url at the first `:` only.

**TODOs**: Come up with a general way of escaping data passed into schemes?

**Fixes**: #645 #644


jackpot51:
Nice job!

Issue #610, opened 2016-04-09
API documentation and manual pages


Ticki:
Documentation
=============

We suck at API docs. Preferably every entry should have a detailed description of the functionality. Currently, most only have a short trivial description, despite not having a trivial functionality.

- [ ] API docs.
  - [x] libextra
  - [x] libmalloc
  - [x] binutils
  - [x] Sodium
  - [ ] OrbTK
  - [ ] Orbclient
  - [ ] Redox-specific std
  - [ ] System
    - [x] Helper functions
    - [ ] Errors
    - [ ] Graphics
    - [ ] Scheme
      - [ ] Scheme
      - [ ] Packet
    - [ ] syscall
      - [ ] Constants
      - [ ] syscalls
        - [ ] sys_debug
        - [ ] sys_supervise
        - [ ] sys_brk
        - [ ] sys_chdir
        - [ ] sys_clock_gettime
        - [ ] sys_clone
        - [ ] sys_close
        - [ ] sys_dup
        - [ ] sys_execve
        - [ ] sys_exit
        - [ ] sys_fpath
        - [ ] sys_fstat
        - [ ] sys_fsync
        - [ ] sys_ftruncate
        - [ ] sys_getpid
        - [ ] sys_iopl
        - [ ] sys_link
        - [ ] sys_lseek
        - [ ] sys_mkdir
        - [ ] sys_nanosleep
        - [ ] sys_open
        - [ ] sys_pipe2
        - [ ] sys_read
        - [ ] sys_rmdir
        - [ ] sys_stat
        - [ ] sys_unlink
        - [ ] sys_waitpid
        - [ ] sys_write
        - [ ] sys_yield
  - [ ] Kernel
    - [x] Top-level modules.
    - [x] Top-level constants.
    - [x] Top-level functions.
    - [ ] ACPI
    - [ ] Alloc
    - [ ] Arch
      - [ ] context
        - [ ] ContextManager
        - [ ] Context
        - [ ] ContextMemory
        - [ ] ContextFile
        - [ ] EnvironmentVariable
        - [ ] ContextZone
        - [ ] context_switch
        - [ ] context_clone
        - [ ] context_box
        - [ ] context_userspace
        - [ ] Constants
      - [ ] elf
      - [ ] Intex et al.
      - [ ] memory
        - [x] Memory
        - [x] MemoryMapEntry
        - [x] address_to_cluster
        - [x] alloc
        - [x] alloc_aligned
        - [ ] alloc_size
[...]

PR #652, opened 2016-06-07
Abstraction, documentation, and uniformity

ticki:
**Main problem**: The Redox code base is in a state in which, it has relatively low code quality. There are multiple reasons for so:

1. Lack of good documentation.

2. Lack of uniformity in naming, code style, and so on.

3. Low use of abstraction.

4. Over use of e.g. heap allocation.

**Solution**: This PR takes a micro step in the right way.

- Follow the naming convention in the `syscall` module (in particular, we throw the `do_sys_` prefix out).
- Make Context names copy-on-write, effectively reducing the allocations on boot by 40-45%.
- Add #[deny(warnings)] to avoid regression.
- Format the code style through a patched version of rustfmt.
- Add a rustfmt config file.

**Drawbacks**: None.

**TODOs**: Lots of things:

1. There are "FIXME"s and "TO-never-DOs" all over the place in the kernel. Many of which are already solved, or really ought to be solved.

2. There are no proper Rustdoc documentation for most of the internals.

3. There are still a long way: We can remove more allocations and avoid more dynamic dispatch.

4. Compile a list of things that needs cleanup, abstraction, or full rewrite.

**Fixes**: No issues.

**State**: Ready.

**Blocking/related**: Related to #610.


ticki:
Considering that the Redox code base is still relatively small and maintainable, I believe that we can actually fix this problem entirely in only a few months. I consider [ralloc](https://github.com/redox-os/ralloc) to be of high code quality, so it can be used as a reference for various tasks.

ca1ek:
Chapter 7.4. of the book, Best Practices should get some more love. It should cover not just the code but also how to use Git, write good commit messages, something I struggle a lot with personally.

ticki:
Agreed.

NilSet:
>This PR takes a micro step in the right way.
>+1,983 −1,573

So a whole step would change 1,983,000,000 lines? :smile:

ca1ek:
Micro step for a programmer, Macro step for the project.

ticki:
There aren't 1,983,000,000 lines in Redox.

jackpot51:
Doing a rustfmt is essentially throwing a smoke grenade on your real changes, it will take me much longer to go through this PR because of it. Perhaps you could undo it and submit it in a separate PR?

ticki:
@jackpot51 I made sure that the rustfmt'ing was only in https://github.com/redox-os/redox/pull/652/commits/da8bd1d885c1e8438c3dd0e37b1cea07bef7283a.

jackpot51:
I am going to be cherry picking what I like, I don't like the format. Not right now, at least

jackpot51:
I also have a big change sitting in another branch that makes debug! and debugln! never allocate, and gets rid of the debug::d* stuff. I will probably rebase these two changesets together

jackpot51:
Then, after that is done, I would like to run the rustfmt

kernel/main.rs
```diff
@@ -5,8 +5,9 @@
            unwind_attributes, zero_one, collections_range, question_mark)]
 #![no_std]
 
-#![allow(deprecated)]
-// #![deny(warnings)]
+// Its me, ur friend. If you ever comment this line, you will end up in hell. Seriously, don't. You
```
jackpot51: I will not bring in this commit. There are good reasons to have warnings at the moment, as we have areas of the code like dup that are not feature complete, and the programmer should be reminded of that every time they build the kernel, but still be able to run the kernel

kernel/syscall/execute.rs
```diff
@@ -212,7 +213,7 @@ pub fn execute(mut args: Vec<String>) -> Result<usize> {
 
                     // debugln!("{}: {}: execute {}", context.pid, context.name, url.string);
 
-                    context.name = url.to_string();
+                    context.name = url.scheme().to_owned().into();
```
jackpot51: This looks odd, this will probably break the context list `ps` because all entries will be named "file"

kernel/main.rs
ticki: You can still have code with warnings. You can locally allow warnings by `#[allow(warnings)]`.

jackpot51:
"Improve performance..." - merged
"Add a rustfmt.toml..." - skipped until merge of debug_cleanup
"Reorder the feature gates" - merged
"Bring back #[deny(warnings)]" - skipped until fix of dup
"Fix CONTRIBUTING.md" - merged
"Make use of COW..." - merged
"Merge" - no longer necessary after previous merges

ticki:
:+1:

Issue #651, opened 2016-05-31
Switch from multirust to rustup


SJFrK:
I'd like to start a motion to move from multirust to [rustup](https://www.rustup.rs/) in bootstrap.sh.

According to the multirust repository it is not maintained anymore and recommends to migrate to rustup, which is an official Rust project.

The change shouldn't be too difficult (at least on Linux and OSX, I don't know how it's handled in Windows), since the syntax is the same as in multirust.

DomThePorcupine:
Not difficult at all, I had no idea it was no longer being maintained. If others agree I'll push the changes today.

ca1ek:
> The change shouldn't be too difficult (at least on Linux and OSX, I don't know how it's handled in Windows)

We don't support Windows at the moment anyways. Or at least I think so.

goolic:
Closed ?

DomThePorcupine:
@goolic not closed. I didn't see all the reactions on my comment, apparently github doesn't send notifications about them. The changes are local on my machine and the pull request will be up later today. I added a couple extras to it as well!

PR #655, opened 2016-06-12
Switch to rustup

DomThePorcupine:
**Problem**: `multirust` has been deprecated, we need to start using `rustup`!

**Solution**: Change the bootstrap file

- Switch from `multirust` to `rustup`
- Add warnings that bootstrap uses `curl | sh` style commands to install things. I feel that I should have been warning people before, these type of commands can make people feel uncomfortable.
- Tried to start better documentation
- Added a new option, you can now run `./bootstrap -s` to see the current travis build status

**Fixes**: #651 

**State**: Ready-ish

**Other**: I'd like some feedback on the `rustInstall` function and how I handle trying to install rustup/uninstall multirust. Another issue involves the $PATH environment variable and making sure it is set properly I can set it in bashrc but i worry that it _might_ not work for people using something like zsh. @jackpot51 I'd appreciate it if a couple of you looked over things before it gets merged.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -1,14 +1,20 @@
 #! /bin/bash
-banner() {
+
+##########################################################
+# This function is simply a banner to introduce the script
+##########################################################
+banner()
+{	
 	echo "|------------------------------------------|"
 	echo "|----- Welcome to the redox bootstrap -----|"
 	echo "|------------------------------------------|"
 }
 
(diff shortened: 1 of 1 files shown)


xliiv:
Shouldn't be [readme.md](https://github.com/redox-os/redox/blob/master/README.md) updated also?

PR #664, opened 2016-07-09
fix fuse pkg name on fedora

stenbock:
**Problem**: 
Fails to build disk image on fedora since bootstrap.sh had the wrong package name for fuse-devel

**Solution**: 
Updated bootstrap.sh for fedora.

**State**:
Ready


DomThePorcupine:
Thanks! I tried to check all these, but it seems as though I missed one. Also welcome to the project!

PR #671, opened 2016-07-22
Improve Stats struct, in fs.rs

k0pernicus:
Add extra functions to obtain the anode, mode and modified date of a given file

**Problem**: No information about the inode, the modified time and the mode of a file/directory.

**Solution**: Add 3 new functions to get from `Stats` the inode, modified time and the mode of the given file/directory.


jackpot51:
We should try to match the Rust libstd, see https://doc.rust-lang.org/std/fs/struct.Metadata.html and https://doc.rust-lang.org/std/os/unix/fs/trait.MetadataExt.html

k0pernicus:
It's ok, I will update the code :-)

jackpot51:
Thanks!

PR #679, opened 2016-07-29
Add SolusOS to the bootstrap script

pwoolcoc:
Just like the title says, allows SolusOS linux users to run the bootstrap script


jackpot51:
cool,  I don't  have solus so I am curious to see other peoples results after merging

PR #683, opened 2016-08-02
The Return of the Unsung Hero

stratact:


kernel/arch/context.rs
```diff
@@ -289,20 +283,16 @@ pub unsafe fn context_clone(regs: &Regs) -> Result<usize> {
...
-                        Err(_err) => () //debugln!("{}: {}: failed to dup resource {} for {}: {}", parent.pid, parent.name, file.fd, clone_pid, err)
-                    }
-                }
+                let files: Vec<ContextFile> = (*parent.files.get())
+                    .iter()
+                    .filter(|file| file.resource.dup().is_ok())
+                    .map(|file|
+                        ContextFile {
+                            fd: file.fd,
+                            resource: file.resource.dup().unwrap(),
```
bjorn3: What if dup succeed the first time and fails the second time?

stratact: I `filter()`'d the iterator beforehand to contain only items that have `Ok(resource)` when doing the dup. So it's safe to unwrap without panicking, if that's what you are referring to?

bjorn3: What if the first dup changes the state in such way that the second dup fails?

ticki:
gr8 m8

PR #684, opened 2016-08-02
Implement duplicate, limit Travis mem usage

NilSet:
This bit of plumbing had been left unimplemented, making try_clone fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ before_install:
       fi
 script:
   - make clean
-  - make all -j
+  - make all -j 3
 notifications:
   email: false
   webhooks: http://37.139.9.28:54863/travis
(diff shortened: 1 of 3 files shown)


NilSet:
After getting frustrated with the travis builds failing intermittently, I added a commit that limits the make jobs to 3. I think the travis builds were being killed for using too much memory (over 10 gb by my count on my local machine) when allowed to run all independent targets at once.

The last complete build on travis took 10 minutes 38 seconds, and for this PR, travis built it in 12 minutes 4 seconds. I think that degree of slowdown is a reasonable price for what I hope is now reliable build status.

While experimenting with parallel builds I spotted a dependency error which I also fixed in this commit.

jackpot51:
Thanks

PR #688, opened 2016-08-04
Fix networking under VirtualBox

haellowyyn:
**Problem**: When building Redox and running it under VirtualBox (`$ make all; make virtualbox`) the DHCP lease seems to fail and there is no network access (`$ ./dl_bmp.sh` hangs). This is reproducible on OS X 10.11.6 and VirtualBox 5.1.2.

**Solution**: In the VM's network adapter settings, setting the NIC's "Cable Connected" option solves the problem.

- In the Makefile target responsible for creating the VirtualBox VM, an additional invocation of `modifyvm` is added to enable 'cableconnected'.


jackpot51:
Thanks!

PR #698, opened 2016-08-19
Use goblin

m4b:
Demonstrates switching to goblin as your ELF parser.  Feel free to not merge (also I'm based off a 2 week old master, sorry), just a suggestion (I've mostly been testing the pure switch and compiling your kernel as it's a good litmus test to make sure no std, fd, io, etc., is present)

Reasons you might want to switch:

1. If in the future you decide to pursue a dynamic linker in rust, as [dryad](http://github.com/m4b/dryad) has much of this work already implemented, and uses the same ELF backend.
2. It's fun
3. The symbol lookup function (while no currently used) isn't the the best way to lookup symbol names, using goblin's utilities would likely help

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -94,3 +94,6 @@
 [submodule "crates/sodium"]
 	path = crates/sodium
 	url = https://github.com/redox-os/sodium.git
+[submodule "crates/goblin"]
+	path = crates/goblin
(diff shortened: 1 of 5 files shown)


jackpot51:
I like `goblin`, looks cool!

ticki:
@m4b Very nice work.

PR #707, opened 2016-09-08
kernel: improve syscall argument handling

pokeylope:
**Problem**: Validation of syscall arguments is inconsistent.

Most syscall handlers correctly validate user-supplied pointers, but some don't, allowing a userspace process to read or modify kernel data.

**Solution**: Instead, do all argument validation in the syscall dispatcher before the specific handler is called. This restricts the usage of raw pointers as much as possible and makes it easier to verify that there is no unsafe behavior. 

**TODOs**: `exec` arguments are still not validated, since it's still using the old C-string ABI and it's more complicated to handle correctly.

**Other**: I saw that there was a kernel rewrite in progress after finishing this, but figured I might as well submit it.


jackpot51:
Looks good, thanks!

PR #713, opened 2016-10-16
Update to upstream feature inversion for goblin 

m4b:
* update to use goblin elf64, elf32 features, etc.
* use program_header constants instead of raw numbers in execute syscall


jackpot51:
Thanks!

PR #721, opened 2016-10-22
Massaging linter with non_upper_case_globals.

Yoric:
**Problem**: Redox won't build with the latest Rust Nightly. See https://github.com/rust-lang/rust/issues/36258.

**Solution**: surgical use of `#[allow(non_upper_case_globals)]`

**TODOs**: Also needs to happen in `libstd`.


Yoric:
Note: Travis build fails because we need to also integrate

- https://github.com/redox-os/ion/pull/190
- https://github.com/redox-os/libstd/pull/1

to fix the build.

PR #726, opened 2016-11-01
Building on macOS

Yoric:
This patch:
- increases the size of the redoxfs image, which was too small to accomodate all the data;
- causes all manipulation of files that belong to `root` to take place in a `sudo`, as macOS would block them otherwise.

Along with
https://github.com/Yoric/redoxfs/commit/0353bfc00c3b9c5351abf32fe60314bc5641165b,
this should fix the ability to build the new kernel on macOS.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -441,19 +441,19 @@ $(BUILD)/filesystem.bin: \
 		filesystem/bin/smith \
 		filesystem/bin/tar
 	rm -rf $@ $(BUILD)/filesystem/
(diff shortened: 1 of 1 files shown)


bjorn3:
This would require people to have root access to build redox. Maybe there is a way to tell redoxfs to make some files owned by root without using fuse.

jackpot51:
Fixed it another way in https://github.com/redox-os/redox/commit/8bbfb8bf2ae89122de8fc0a2d5b3d6f52307bca0

PR #729, opened 2016-11-02
increase process id range

pythoneer:
**Problem**: The current number of possible process ids is very small and could easily lead to a reuse of those.

**Solution**: increase the range to usize::max_value() - 1

CONTEXT_MAX_CONTEXTS

**Drawbacks**: the number pids is limited to usize maybe use u64 instead?

**TODOs**: nothing

**Fixes**: nothing

**State**: WIP

**Blocking/related**: nothing

### DIFF
--- a/kernel/context/mod.rs
+++ b/kernel/context/mod.rs
@@ -26,7 +26,7 @@ pub mod file;
 pub mod memory;
 
 /// Limit on number of contexts
-pub const CONTEXT_MAX_CONTEXTS: usize = 65536;
+pub const CONTEXT_MAX_CONTEXTS: usize = usize::max_value() - 1;
 
 /// Maximum context files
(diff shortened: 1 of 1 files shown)


kernel/context/mod.rs
```diff
@@ -26,7 +26,7 @@ pub mod file;
 pub mod memory;
 
 /// Limit on number of contexts
-pub const CONTEXT_MAX_CONTEXTS: usize = 65536;
+pub const CONTEXT_MAX_CONTEXTS: usize = usize::max_value();;
```
jackpot51: Extra semicolon. Also, I think it should be usize::max_value() - 1 so that it doesn't cause an overflow panic

jackpot51:
Thanks!

PR #732, opened 2016-11-03
Improving self-documentation of the implementation of Registry.

Yoric:
Replacing instances of `(usize, usize)` by records with meaningful field names. This should make the code a bit easier to read.


jackpot51:
Thanks for that!

PR #735, opened 2016-11-03
Update README.md

msehnout:
Current version does not work. Using relative links as described here: https://help.github.com/articles/relative-links-in-readmes/


msehnout:
Oh, seems to fixed already.

PR #737, opened 2016-11-04
Minor improvements on schemes.

Yoric:
Replace `usize` with `SchemeId` wherever applicable. This should remove a few footguns.


Yoric:
Test fails with
```
error[E0523]: found two different crates with name `syscall` that are not distinguished by differing `-C metadata`. This will result in symbol conflicts between the two.

 --> programs/orbutils/src/terminal/main.rs:6:1

  |

6 | extern crate syscall;

  | ^^^^^^^^^^^^^^^^^^^^^
```

I don't think it's my fault.

Yoric:
New version coming.

PR #740, opened 2016-11-04
bootstrap.sh: Make use of yum instead of dnf on Fedora

genodeftest:
Fully adapt this script to using dnf instead of yum on Fedora.

**Problem**:
Fedora ships dnf in all supported versions replacing yum in default installations. If a user doesn't have yum installed, detection of Fedora and installation of packages might fail.

**Solution**:
dnf is mostly a drop-in replacement for yum so you can just swap command names

- replace usage of yum through dnf
- update detection of fedora installation to use dnf instead of yum

**Drawbacks**:
Old (outdated and unsupported) installations of Fedora will not be able to run this script correctly any more. This is true for Fedora ≤ 21, Fedora 22 was released 2.5 years ago. Some ultra-conservative users keep using yum instead of dnf, they will see this script failing but probably won't be using "bleeding-edge" technology like Rust anyway.

**TODOs**:
None

**Fixes**:
None

**State**:
ready to be merged

**Blocking/related**:
None

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -134,19 +134,19 @@ fedora()
 	echo "Detected Fedora"
 	if [ -z "$(which git)" ]; then
 		echo "Installing git..."
-		sudo yum install git-all
+		sudo dnf install git-all
 	fi
 	if [ "$1" == "qemu" ]; then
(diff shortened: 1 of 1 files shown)


jackpot51:
Thanks

PR #749, opened 2016-11-06
add additional dependencies for MacOS users

GlenDC:
**Problem**: Currently the install/update dependencies description seemed to be only targeted at Linux Users.

**Solution**: In This PR I separated out the dependency that's for Linux Users only, and added extra dependencies for MacOS Users.


jackpot51:
Thanks. Isn't it available on homebrew by tapping this: https://github.com/sevki/homebrew-gcc_cross_compilers/blob/master/x86_64-elf-gdb.rb?

PR #751, opened 2016-11-06
update dependencies README

GlenDC:
@jackpot51 I updated the README instructions, to make it clearer for the 3 different target groups: Linux Users, MacOS-MacPorts Users and MacOS-Hombrew users.

The steps are now pretty complete and made a similar modification to the redox-os/book @ https://github.com/redox-os/book/pull/61


jackpot51:
very nice

PR #754, opened 2016-11-14
Making file handles, context ids, scheme ids type-safer and better self-documented.

Yoric:
**Problem**: We're using `usize` for file descriptors, pids, scheme ids, ... It works, but it's also non-documenting and somewhat footgunish.

**Solution**: This PR introduces types `FileHandle`, `ContextId`, `SchemeId`, their atomic counterparts and a macro to generate similar types. At runtime, they behave exactly like `usize` (respectively `AtomicUsize`), but with added type-safety. This patch covers only the kernel.

**State**: Ready.

**Blockers**: Not strictly a blocker, but I expect to build upon this for https://github.com/redox-os/rfcs/pull/2.

### DIFF
--- a/kernel/common/int_like.rs
+++ b/kernel/common/int_like.rs
@@ -1,0 +1,108 @@
+//! Helpers used to define types that are backed by integers (typically `usize`),
+//! without compromising safety.
+//!
+//! # Example
+//!
+//! ```
+//! /// Define an opaque type `Pid` backed by a `usize`.
+//! int_like!(Pid, usize);
+//!
+//! const ZERO: Pid = Pid::from(0);
+//! ```
+//!
+//! # Example
+//!
+//! ```
+//! /// Define opaque types `Pid` and `AtomicPid`, backed respectively by a `usize`
+//! /// and a `AtomicUsize`.
+//!
+//! int_like!(Pid, AtomicPid, usize, AtomicUsize);
+//!
+//! const ZERO: Pid = Pid::from(0);
+//! let ATOMIC_PID: AtomicPid = AtomicPid::default();
+//! ```
+
+#[macro_export]
+macro_rules! int_like {
+    ($new_type_name:ident, $backing_type: ident) => {
+        #[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Clone, Copy)]
+        pub struct $new_type_name($backing_type);
+
+        impl $new_type_name {
+            pub const fn into(self) -> $backing_type {
+                self.0
+            }
+            pub const fn from(x: $backing_type) -> Self {
+                $new_type_name(x)
+            }
+        }
+    };
+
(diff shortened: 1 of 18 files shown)


Yoric:
Pinging @jackpot51 for review.

kernel/common/int_like.rs
```diff
@@ -0,0 +1,108 @@
...
+            pub const fn into(self) -> $backing_type {
+                self.0
+            }
+            pub const fn from(x: $backing_type) -> Self {
+                $new_type_name(x)
+            }
+        }
+    };
+
+    ($new_type_name:ident, $new_atomic_type_name: ident, $backing_type:ident, $backing_atomic_type:ident) => {
```
NilSet: This feels awkward to stick it on the end of another type def, shouldn't atomic_int_like be its own macro?

kernel/common/int_like.rs
```diff
@@ -0,0 +1,108 @@
...
+//! ```
+
+#[macro_export]
+macro_rules! int_like {
+    ($new_type_name:ident, $backing_type: ident) => {
+        #[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Clone, Copy)]
+        pub struct $new_type_name($backing_type);
+
+        impl $new_type_name {
+            pub const fn into(self) -> $backing_type {
```
NilSet: Why not impl [From](https://doc.rust-lang.org/core/convert/trait.From.html) (and get Into for free)

kernel/common/int_like.rs
```diff
@@ -0,0 +1,108 @@
...
+        impl $new_atomic_type_name {
+            pub const fn new(x: $new_type_name) -> Self {
+                $new_atomic_type_name {
+                    container: $backing_atomic_type::new(x.into())
+                }
+            }
+            pub const fn default() -> Self {
+                Self::new($new_type_name::from(0))
+            }
+            pub fn load(&self, order: ::core::sync::atomic::Ordering) -> $new_type_name {
```
NilSet: Is there a chance this wrapper code makes the operation not quite so atomic?

kernel/common/int_like.rs
Yoric: Because `From` and `Into` are not `const` and `const` is needed in a few places of the existing code, to initialize the atomics.

kernel/common/int_like.rs
Yoric: How so? The memory access is still atomic, and if my memory serves, atomic operations guarantee only that.

kernel/common/int_like.rs
NilSet: You could initialize with a `const fn new` like you are doing for the atomic one below.

Yoric: Sure, I could replace `from` with `new`. And truth be told, I probably don't need `into` to be `const`.

If that's better, I can update the patches to do that.

Issue #720, opened 2016-10-22
Support MacPorts in bootstrap.sh


Yoric:
For the moment, on Mac, bootstrap.sh assumes that the user uses Brew. Ideally, it should also support MacPorts.

Yoric:
Documenting the list of ports that I have installed so far:

- gmake
- qemu
- nasm
- osxfuse
- gcc49

Yoric:
The above list seems to do the trick.

DomThePorcupine:
@Yoric are you proposing one of us do this, or would you like to work on it?

Yoric:
Well, I admit I hoped that someone else would do it :)

But if needed, I guess I can do it. It's a bit tricky to test, though.

DomThePorcupine:
That's fine, I'll start working on it. When I am done maybe you can help test it.

Yoric:
Apparently, with the new kernel, we also need `x86_64-elf-gcc` (probably instead of `i386-elf-gcc`).

GlenDC:
@DomThePorcupine are you still working on this task?

Otherwise I might want take it over. As I also just updated the documentation regarding this (more or less) in the redox-os/redox/README.md file and redox-os/book.

@Yoric I think it could be useful to allow the user to JUST install the dependencies, and skip the entire boot part. Would it be fine if I provide a CL_OPT that allows you to skip the boot process?

Yoric:
@GlenDC Yes, I personally haven't used bootstrap.sh at all, as it didn't work for MacPorts. I think it would be a good idea to let users skip it if dependencies are installed. Perhaps it should only print the list of dependencies?

DomThePorcupine:
@GlenDC if you want go for it i won't get to it for a while

GlenDC:
Ok today I was away all day, but tomorrow I'll work on a solution and reference this issue so you can give your input on my proposed PR regarding this issue.

GlenDC:
@Yoric and @DomThePorcupine the PR mentioned yesterday has been opened, feel free to give your input at PR #753

jackpot51:
Looks good to me, it would be good to have some OS X users with different setups test it.

GlenDC:
I tested it both with `MacPorts` and `Homebrew`, but you're right @jackpot51 , probably better if some more MacOS people (with different setups) can confirm.

meiyopeng:
@GlenDC 
1. Please fix this:
[...]

PR #759, opened 2016-11-25
fix pkgconfig darwin macports pkg in bootstrap.sh

GlenDC:
**Problem**: bootstrap.sh tried to download `pkg-config` when using Macports, even though it is called `pkgconfig` as found here: https://www.macports.org/ports.php?by=name&substr=pkgconfig

**Solution**: Replace `pkg-config` with `pkgconfig`.


GlenDC:
This fixes the last problem reported by @meiyopeng in issue #720.

PR #761, opened 2016-11-26
Fix building with the latest cargo nightly

NilSet:
This commit by @shawnanastasio fixes changes introduced by nightly cargo which are currently breaking our build.


jackpot51:
This *must* be rewritten as a shell script - it should be easy to do and will remove a dependency on perl

NilSet:
Will do.

jackpot51:
Thanks

PR #763, opened 2016-11-27
Upgrade rand to 0.3.15

aesedepece:
After https://github.com/rust-lang-nursery/rand/commit/f2267c1d03b22349d7dedc1d823ebd115f820341, there is no `0.3.14` release for https://github.com/rust-lang-nursery/rand .

This commit fixes compilation of redox by upgrading rand's version to `0.3.15`.


jackpot51:
we can remove the replace sections now that it was bumped - I will take care of that.

Thanks

PR #785, opened 2016-12-15
Fix bug, mouse and keybroad no response in virtualbox

origel:
**Problem**: [Sometimes, when Redox startup, the mouse and keybroad is no response in virtualbox.]

**Solution**: [make nestedpaging param off.]


jackpot51:
Thanks!

PR #794, opened 2016-12-17
Context: truncate files vector

jaje:
**Problem**: Size of context files vector is never reduced.

**Solution**: Adjust files vector to smallest size if possible.

- Truncate files vector length if possible.
- Shrink files vector capacity if possible.

**Drawbacks**: If process opens and closes lots of files, this change may increase count of relocation.

**Fixes**: no related issue

**State**: ready

### DIFF
--- a/kernel/context/context.rs
+++ b/kernel/context/context.rs
@@ -221,11 +221,21 @@ impl Context {
     }
 
     /// Remove a file
-    // TODO: adjust files vector to smaller size if possible
     pub fn remove_file(&self, i: FileHandle) -> Option<File> {
         let mut files = self.files.lock();
         if i.into() < files.len() {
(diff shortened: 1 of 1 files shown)


jackpot51:
I am a bit split on this - it takes a lot of cpu time to shrink the vec compared to leaving empty alots, and it does not save much heap memory - maybe 32 bytes per file handle.

jaje:
Yes, that is true, but it is not generally good idea to keep unused memory when not needed.
Maybe it would be better to shrink it only when there is more empty memory. 
______
_10 is just example, I do not think that there is enough programs to determine the optimal value_

jackpot51:
This would need to be moved to the kernel repository.

PR #796, opened 2016-12-20
add chmod command support

origel:
**Problem**: [add chmod func for init.]

**Solution**: [add chmod func for init.]


jackpot51:
What is the purpose? The init scripts do not need to run chmod, and chmod already exists in coreutils

jackpot51:
If you want something to work on - you can look at https://github.com/redox-os/coreutils and add missing commands there

origel:
OK

PR #798, opened 2016-12-28
Fix vesad text rendering

xTibor:
This pull request fixes a small bug in the bitmapped font rendering routine which caused the glyphs to be rendered incorrectly. It was really visible around the block drawing characters.

Screenshots:
[Before the fix](http://i.imgur.com/mqrJnth.jpg) - [After](http://i.imgur.com/AgTEHvP.jpg)


jackpot51:
Thanks!

PR #801, opened 2017-01-01
moved "old kernel" and "ZFS" at the bottom of the software list

FaraoneLele:
moved "old kernel" and "ZFS" at the bottom of the software list, because they are abandoned and the 'interest is smaller than the other


jackpot51:
Cool, thanks!

FaraoneLele:
What about putting the software in alphabetic order?

jackpot51:
Sure, that would be good.

PR #807, opened 2017-01-02
Implementing sys_dup_from/sys_dup_to.

Yoric:
Requires https://github.com/redox-os/syscall/pull/1


Yoric:
Wait, that implementation is wrong. I'll upload a fix later.

jackpot51:
Thanks for this!

Issue #790, opened 2016-12-16
Implement dup_from/dup_to


Yoric:
These operations are detailed in the RFC on [userspace capabilities](https://github.com/redox-os/rfcs/pull/2).

@jackpot51 Could you assign this issue to me?

Yoric:
Note that there is a large intersection between `dup_from` and `open`. Both syscalls may end up merged.

Yoric:
A few updates to this issue.

**Userspace**

The initial idea was to delegate most of the implementation of `dup_from`/`dup_export` to userspace. This has the following consequences:

1. each call to `dup_from` and to `dup_export` needs to specify the path (or at least the scheme), which is awkward;
2. basically all userspace implementations of `dup_from`/`dup_export` would be the same, which is feasible but a bit annoying to achieve.

For the time being, we decided to avoid this. We might return here at some point, who knows.

**ContextId**

The initial idea was that `dup_export` could be used to send a file descriptor to a specific process based on its pid, while `dup_from` could be used to receive a file descriptor from a specific process based on its pid. As it turns out, that's very prone to race conditions (what happens if one of the processes dies and is replaced with another process with the same pid?) which can be exploited for nefarious purposes.

Consequently, using the pid is a bad idea.

Back to the drawing board.

Yoric:
Ok, I'm back from the drawing board. Here's an idea. @jackpot51 , @ticki, any thoughts?

# Objective

Provide a way for a process to open a file descriptor and send it to another process, without relying upon pids (see [above](https://github.com/redox-os/redox/issues/790#issuecomment-272120475).

# Cables

At high-level, instead of using syscalls to send/receive file descriptors, we'll use a mechanism of dedicated pipes. Since words "pipes" and "channels" are already taken, we'll need another name, say "cables". A `cable` is used exclusively to send/receive pairs of `(file descriptor, data)`, where `data` is typically a key which may be used for (de)multiplexing file descriptors.

A `cable` is created by opening the `cable:` scheme and is shared with the usual rules for sharing file descriptors (i.e. it's duped during calls to `fork`). A `cable` supports the usual `read`/`write` operations, with the following semantics:

- `write()` must always be called to write a `usize`-sized file descriptor followed by an optional `[u8; ...]` message;
- calling `write()` causes the file descriptor to be `dup()`ed';
- messages are received in the order in which they are sent;
[...]

PR #810, opened 2017-01-02
Issue #790

Yoric:
Ok, this one should assign a meaningful number to file descriptors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ clean:
 	cargo clean --manifest-path schemes/redoxfs/Cargo.toml
 	cargo clean --manifest-path schemes/tcpd/Cargo.toml
 	cargo clean --manifest-path schemes/udpd/Cargo.toml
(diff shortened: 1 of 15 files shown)


Yoric:
Mmmh... the build failure seems to be in C code. Not sure there's anything I can do about it.

Yoric:
This is a full reimplementation, based on our earlier discussion. `dup_from`/`dup_export` is now implemented fully in the kernel and uses a `key` instead of a path to identify file descriptors.

It's still missing one pretty large bit: cleaning up file descriptors exported to a process `pid` when `pid` goes down. I haven't found where in the code we implement processes going down, but I'll try and do this in the next few days.

Yoric:
Oh, and of course, there is the glaring absence of any test. @jackpot51, where should I put tests for this kind of feature?

jackpot51:
This has conflicts due that need to be resolved

PR #820, opened 2017-01-07
fix bootstrap issue on archlinux 64 bit

ZeppLu:
On Archlinux 64 bit, `qemu-system-i386' is in "qemu-arch-extra" package instead of "qemu"

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -156,7 +156,11 @@ archLinux()
 	if [ "$1" == "qemu" ]; then
 		if [ -z "$(which qemu-system-i386)" ]; then
 			echo "Installing QEMU..."
-			sudo pacman -S qemu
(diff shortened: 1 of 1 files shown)


jackpot51:
We don't use qemu-system-i386, we use qemu-system-x86_64

Dohxis:
Everything runs fine on mine Arch Linux machine with `qemu`.

ZeppLu:
I haven't got time to compile and run it yet, I mean it's something wrong in `bootstrap.sh`. If redox os does not depend on `qemu-system-i386`, in that script you should check `qemu` instead.

jackpot51:
The bootstrap no longer checks for qemu-system-i386, it checks for qemu-system-x86_64

PR #845, opened 2017-02-13
use smoltcp as networking stack

little-dude:
Hi @jackpot51,

Here is my take at replacing the existing network stack by [smoltcp](https://github.com/m-labs/smoltcp).

Rationale
=========

Benefits from using a third-party network stack:

- Less code to maintain: `smoltcp` is currenlty written and maintained by @whitequark.
- If `smoltcp` get adopted by many project, we can reasonably expect it to grow in terms of features, code quality, and performance. For instance it might get [IPv6](https://github.com/m-labs/smoltcp/issues/3) soon.

Benefits from using `smoltcp`:

- It's built in such a way that it is theoritically possible for the whole network stack not to allocate any memory. This could be highly desirable, to run redox on systems with very little memory, or simly to offer a network stack with a very low latency.
 - The code is easy to read and well documented. As a beginner in system programming I learnt a lot by reading the code and the [various](https://lab.whitequark.org/notes/2016-12-17/owning-collections-in-heap-less-rust/) [blog](https://lab.whitequark.org/notes/2017-01-16/abstracting-over-mutability-in-rust-macros/) [posts](https://lab.whitequark.org/notes/2016-12-13/abstracting-over-mutability-in-rust/) about smoltcp implementation.

Disclaimer
==========

I'm a newbie in system programming, and this is my first contribution to Redox.  I did my best to make the code as correct as possible but there may be obvious mistakes.

The implementation is not complete yet but I'm making this PR to get early feedback. Basically, am I in the right direction? Are there any fundamental design flaws?

Implementation details
======================

- I removed `ipd`, `udpd` and `tcpd`. Everything is contained in `ethernetd`
- TCP sockets are not yet implemented.
- I introduced to new scheme methods: `sockrecv` and `socksend`, because I could not find a way to implement UDP sockets with only `read` and `write`.  The problem is that the [`std::net::UdpSocket::recv_from()`](https://doc.rust-lang.org/std/net/struct.UdpSocket.html#method.recv_from) and [`std::net::UdpSocket::send_to()`](https://doc.rust-lang.org/std/net/struct.UdpSocket.html#method.send_to) semantics requires an extra endpoint argument to be passed to the method that read/write to/from the socket. Basically, I needed something similar to `sendto` and `recvfrom` on Linux.
- As said previously, the implementation is incomplete.


little-dude:
This builds but requires https://github.com/redox-os/syscall/pull/2 and some hacks in smoltcp: https://github.com/little-dude/smoltcp/commit/2d5deaf8e0e6bbab0d8374f6be38d09fbca65252

whitequark:
> It's built in such a way that it is theoritically possible for the whole network stack not to allocate any memory. This could be highly desirable, to run redox on systems with very little memory, or simly to offer a network stack with a very low latency.

Let me clarify this a bit. Using smoltcp with *no heap allocation whatsoever* requires some ownership gymnastics that really depend on the exact structure of your application. E.g. if you know you only ever need one server socket at port 80, sure. Using smoltcp in an application like redox will require some allocation per socket. However, it is up-front; even in redox, smoltcp has a guaranteed upper latency bound on every operation, or could be easily restructured to be so (depending on how you define the bound).

little-dude:
@jackpot51 UDP is now working, and I'm not using any extra syscall (just `read`, `write`, `fpath` and `dup`).

DHCP is broken, because smoltcp does not support sending UDP packets to non-unicast addresses, so I think I'll need to implemented a DHCP client with an L2 socket directly.

However, we can now configure multiple ip addresses manually:

```
ion:file:/root# ip show
No IP configured on the device
ion:file:/root# ip add 192.168.10.2
Ip address set
ion:file:/root# ip show
Current IPs:
192.168.10.2
```

Also, thanks to smoltcp, we can ping redox:

```
~/r/e/t/debug ❯❯❯ ping 192.168.10.2                           
PING 192.168.10.2 (192.168.10.2) 56(84) bytes of data.
64 bytes from 192.168.10.2: icmp_seq=1 ttl=64 time=1.08 ms
^C
--- 192.168.10.2 ping statistics ---
1 packets transmitted, 1 received, 0% packet loss, time 0ms
rtt min/avg/max/mdev = 1.083/1.083/1.083/0.000 ms
```

The `ip` command is here: https://github.com/redox-os/netutils/pull/18

UDP is working. I added a small `echoserver` command that starts a UDP server that echoes back whatever you send to it. For example, it answers this client:

```rust
use std::env;
use std::net::UdpSocket;

pub fn run() -> Result<(), String> {
    let mut args = env::args().skip(1);
    let local_endpoint = args.next().ok_or("Missing argument".to_owned())?;
    let remote_endpoint = args.next().ok_or("Missing argument".to_owned())?;
    let mut socket = UdpSocket::bind(local_endpoint).map_err(|e| format!("Failed create UDP socket: {}", e))?;
    socket.connect(remote_endpoint).map_err(|e| format!("faile dto connect to remote endpoint {}", e))?;

    let mut buf: Vec<u8> = Vec::new();
    for i in 0..1000 {
        buf.push((i % 255) as u8);
    }

    let mut rx_buf: Vec<u8> = vec![];
    rx_buf.resize(1000, 0);

    for req in 0..1000 {
        let tx_count = socket.send(&buf[..(req + 1) as usize]).map_err(|e| format!("Failed to send data: {}", e))?;
        println!("Send {} bytes", tx_count);
        let rx_count = socket.recv(&mut rx_buf).map_err(|e| format!("Failed to receive data: {}", e))?;
        println!("Received {} bytes", rx_count);
    }
    return Ok(());
}

fn main() {
    match run() {
        Ok(_) => {
            println!("Exiting...");
        },
        Err(e) => {
            println!("Error while running udp echo client: {}", e);
        }
    }
}
```

I have one big problem: I have no idea how to implement read and write timeouts. Currently, the socket is blocking, and will block forever if there is nothing to read.

whitequark:
> DHCP is broken, because smoltcp does not support sending UDP packets to non-unicast addresses, so I think I'll need to implemented a DHCP client with an L2 socket directly.

Uh, this should be fixed in smoltcp, I'm pretty sure.

ids1024:
@little-dude I've modified this to go in the new netstack repo, where the network stack has been moved (https://github.com/ids1024/netstack/tree/smoltcp). How can I test this in qemu?

I'm hoping smoltcp will help with issues I've had getting git/libgit2 to run under Redox; although that will require tcp.

little-dude:
Nice, I'm happy to see someone picking this up.
Wrt testing this, here is how I was doing:

I created a bridge `qemu-br1`:
```sh
ip link add name qemu-br1 type bridge
ip link set qemu-br1 up
ip address add 192.168.10.1/24 dev qemu-br1
```
I created a file `/etc/qemu/bridge.conf` with this content:

```
allow qemu-br1
```
Now you should be able to start redox with a NIC attached to this bridge:
```
SDL_VIDEO_X11_DGAMOUSE=0 qemu-system-x86_64 -serial mon:stdio -d cpu_reset -d guest_errors -smp 4 -m 1024 -machine q35 -nographic -vga none -device nec-usb-xhci,id=xhci -device usb-tablet,bus=xhci.0 -enable-kvm -cpu host -drive file=build/harddrive.bin,format=raw -netdev bridge,br=qemu-br1,id=net1 -device e1000,netdev=net1,id=nic1
```
Now, you may want the bridge to act as a DHCP and/or DNS server. I used `dnsmasq` for that.
Here is my `/etc/qemu/dnsmasq.conf`:

```
# Configuration file for dnsmasq.
#
# Format is one option per line, legal options are the same
# as the long options legal on the command line. See
# "/usr/sbin/dnsmasq --help" or "man 8 dnsmasq" for details.

# Listen on this specific port instead of the standard DNS port
# (53). Setting this to zero completely disables DNS function,
# leaving only DHCP and/or TFTP.
#port=5353

# The following two options make you a better netizen, since they
# tell dnsmasq to filter out queries which the public DNS cannot
# answer, and which load the servers (especially the root servers)
# unnecessarily. If you have a dial-on-demand link they also stop
# these requests from bringing up the link unnecessarily.

# Never forward plain names (without a dot or domain part)
#domain-needed
# Never forward addresses in the non-routed address spaces.
#bogus-priv

# Uncomment these to enable DNSSEC validation and caching:
# (Requires dnsmasq to be built with DNSSEC option.)
#conf-file=/usr/share/dnsmasq/trust-anchors.conf
#dnssec

# Replies which are not DNSSEC signed may be legitimate, because the domain
# is unsigned, or may be forgeries. Setting this option tells dnsmasq to
# check that an unsigned reply is OK, by finding a secure proof that a DS 
# record somewhere between the root and the domain does not exist. 
# The cost of setting this is that even queries in unsigned domains will need
# one or more extra DNS queries to verify.
#dnssec-check-unsigned

# Uncomment this to filter useless windows-originated DNS requests
# which can trigger dial-on-demand links needlessly.
# Note that (amongst other things) this blocks all SRV requests,
# so don't use it if you use eg Kerberos, SIP, XMMP or Google-talk.
# This option only affects forwarding, SRV records originating for
# dnsmasq (via srv-host= lines) are not suppressed by it.
#filterwin2k

# Change this line if you want dns to get its upstream servers from
# somewhere other that /etc/resolv.conf
#resolv-file=

# By  default,  dnsmasq  will  send queries to any of the upstream
# servers it knows about and tries to favour servers to are  known
# to  be  up.  Uncommenting this forces dnsmasq to try each query
# with  each  server  strictly  in  the  order  they   appear   in
# /etc/resolv.conf
#strict-order

# If you don't want dnsmasq to read /etc/resolv.conf or any other
# file, getting its servers from this file instead (see below), then
# uncomment this.
#no-resolv

# If you don't want dnsmasq to poll /etc/resolv.conf or other resolv
# files for changes and re-read them then uncomment this.
#no-poll

# Add other name servers here, with domain specs if they are for
# non-public domains.
#server=/localnet/192.168.0.1

# Example of routing PTR queries to nameservers: this will send all
# address->name queries for 192.168.3/24 to nameserver 10.1.2.3
#server=/3.168.192.in-addr.arpa/10.1.2.3

# Add local-only domains here, queries in these domains are answered
# from /etc/hosts or DHCP only.
#local=/localnet/

# Add domains which you want to force to an IP address here.
# The example below send any host in double-click.net to a local
# web-server.
#address=/double-click.net/127.0.0.1

# --address (and --server) work with IPv6 addresses too.
#address=/www.thekelleys.org.uk/fe80::20d:60ff:fe36:f83

# Add the IPs of all queries to yahoo.com, google.com, and their
# subdomains to the vpn and search ipsets:
#ipset=/yahoo.com/google.com/vpn,search

# You can control how dnsmasq talks to a server: this forces
# queries to 10.1.2.3 to be routed via eth1
# server=10.1.2.3@eth1

# and this sets the source (ie local) address used to talk to
# 10.1.2.3 to 192.168.1.1 port 55 (there must be a interface with that
# IP on the machine, obviously).
# server=10.1.2.3@192.168.1.1#55

# If you want dnsmasq to change uid and gid to something other
# than the default, edit the following lines.
#user=
#group=

# If you want dnsmasq to listen for DHCP and DNS requests only on
# specified interfaces (and the loopback) give the name of the
# interface (eg eth0) here.
# Repeat the line for more than one interface.
interface=qemu-br1

# Or you can specify which interface _not_ to listen on
#except-interface=
# Or which to listen on by address (remember to include 127.0.0.1 if
# you use this.)
#listen-address=
# If you want dnsmasq to provide only DNS service on an interface,
# configure it as shown above, and then use the following line to
# disable DHCP and TFTP on it.
#no-dhcp-interface=

# On systems which support it, dnsmasq binds the wildcard address,
# even when it is listening on only some interfaces. It then discards
# requests that it shouldn't reply to. This has the advantage of
# working even when interfaces come and go and change address. If you
# want dnsmasq to really bind only the interfaces it is listening on,
# uncomment this option. About the only time you may need this is when
# running another nameserver on the same machine.
bind-interfaces

# If you don't want dnsmasq to read /etc/hosts, uncomment the
# following line.
#no-hosts
# or if you want it to read another file, as well as /etc/hosts, use
# this.
#addn-hosts=/etc/banner_add_hosts

# Set this (and domain: see below) if you want to have a domain
# automatically added to simple names in a hosts-file.
expand-hosts

# Set the domain for dnsmasq. this is optional, but if it is set, it
# does the following things.
# 1) Allows DHCP hosts to have fully qualified domain names, as long
#     as the domain part matches this setting.
# 2) Sets the "domain" DHCP option thereby potentially setting the
#    domain of all systems configured by DHCP
# 3) Provides the domain part for "expand-hosts"
#domain=thekelleys.org.uk

# Set a different domain for a particular subnet
domain=qemu-1,192.168.10.0/24

# Uncomment this to enable the integrated DHCP server, you need
# to supply the range of addresses available for lease and optionally
# a lease time. If you have more than one network, you will need to
# repeat this for each network on which you want to supply DHCP
# service.
dhcp-range=192.168.10.2,192.168.10.254,12h
```

You can now run `dnsmasq` with: `/usr/bin/dnsmasq -k --conf-file=/etc/qemu/dnsmasq.conf`

little-dude:
With the instructions above, you can capture traffic directly on `qemu-br1`, and run some tests with tools like `nping` or `iperf`.

This is a little bit old so I'm not sure these instructions are 100% complete, so let me know if you have questions

whitequark:
Heads up, @batonius is doing work on smoltcp for redox concurrently with you.

jackpot51:
Closed, as significant work has been done on the current netstack to switch to smoltcp by @batonius

whitequark:
Wait, is that actually merged?

jackpot51:
It will be soon. See https://github.com/batonius/netstack/tree/smolnetd

It has two problema left to be addressed:
- It panics from remote behavior due to unwrapping the poll function
- Performance needs to be improved

Issue #844, opened 2017-02-10
Use rst for requirements tracking


jackpot51:
Track kernel, driver, scheme, etc. requirements using rst and improve automated tests for compliance

vitiral:
I just saw this. 

To keep everyone up to date, I have been pushing hard to get rst to 1.0. The first part of that was renaming it. 

It is now called artifact and it's project page is here: https://github.com/vitiral/Artifact

One of the major accomplishments this past week was static site generation. You can see the project's own requirements here:

http://vitiral.github.io/artifact/#artifacts/REQ-1

I will be done with pre 1.0 by the end of the week and want to start writing the high level design docs after that.

vitiral:
@jackpot51 can you change the name to "Use artifact" instead of "Use rst"?

PR #856, opened 2017-02-24
initial design docs

vitiral:
**Problem**: write design docs for redox  

**Solution**: this is a first pass at developing design docs and is not in a state to be merged. The purpose of this PR is to get feedback on the overall methodology

- adding a new `design/` folder and `.art` folder with new design documents.

**Drawbacks**: no draw backs

**TODOs**: pretty much everything. The plan is:

- get the high level docs written to a point that they are useful and merge them
- write lowever level docs in separate (smaller) merges and link them to implementation and tests

**Fixes**: #844

**State**: WIP

**Blocking/related**: none


vitiral:
I added an index.html to show the power of exporting the design docs. If you clone and run `firefox index.html` you can see the documents rendered.

vitiral:
As it states in `REQ-0-assertions` I am using the following terminology:

Assertions shall be used throughout these documents to mean:
- **shall:** the statement must be implemented and it's
    implementation verified.
- **will:** statement of fact, not subject to verification.
    i.e. "The X system will have timing as defined in ICD 1234"
- **should:** goals, non-mandatory provisions. Statements using "should"
    **should** be verified if possible, but verification is not mandatory if
    not possible. Is a statement of intent.

Thetoxicarcade:
Fascinating. How is this different from, say, a separate org repo under "redox-os/design"? (which could be a submodule in the operating repository, sure)

vitiral:
we could definitely create a separate org repo. I am totally open to suggestions around that.

You would still want the `.art` config file to be in the root branch though, since you would want to link all the artifacts to source code in the various repos.

jackpot51:
Yes, I think this should be a submodule. I have created the repo here: https://github.com/redox-os/design, and I gave you write access

vitiral:
Awesome. I guess I'll have to fix that bug!

Thetoxicarcade:
If you do that, what url will you want to use? Will that correlate to how the redox-os project handles webpages already?

vitiral:
The url will be `http://redox.github.io/design`

I don't intend for it to be a permanant url (that will be up to @jackpot51) but rather one that we can use while we are developing the initial docs.

bjorn3:
*redox**-os**.github.io

vitiral:
In general, how do people like what I am doing so far? Does everyone like the langauge (**shall**, **will** and **should**)? Does it make sense to translate the book into high level requirements documents?

If so, this pull request should be closed and I will do the initial work from the subrepo until I can't do anything myself anymore (because I don't know enough about redox). Then I'll help others write and review the more detailed design/test docs.

Thetoxicarcade:
So long as the design provides a human readable explanation of how each of the parts works, and links to other parts (This should happen at the code level regardless). I really would love a doc-set that's like @intermezzOS's explanations, but not at the expense of how redox will arc in the future.

vitiral:
What do you think of artifact's own design docs? It will be a lot like these

http://vitiral.github.io/artifact/#artifacts/req-1

Thetoxicarcade:
Implemented / tested are good figures for each part of the specs, or any part of the planned codebase. Do we have a separate spec/design spot that's more approachable?

Thetoxicarcade:
@jackpot51 / @ticki are you good with signing off on the design repo format (as he described) enough to close this issue?

vitiral:
jackpot reviewed the docs and gave his blessing. Further development will happen in the [design](https://github.com/redox-os/design) repo

Issue #850, opened 2017-02-23
Build error


adrianN:
**Reproduction**: Follow the instructions in the book to build redox

**Expected behavior**: Redox is built.

**Actual behavior**: Make fails. See https://gist.github.com/adrianN/2b1c09d83bae4b88acf82a25f8856a0f

**Build information**: [output of `rustc -V`, `git rev-parse HEAD`, `qemu-i386 -version`, `uname -a`, etc.]
MacBook:redox aneumann$ rustc -V
rustc 1.17.0-nightly (306035c21 2017-02-18)
MacBook:redox aneumann$ git rev-parse HEAD
11bc60d3ca2cd9214a0b6e03e1727251f297f15f
MacBook:redox aneumann$ uname -a
Darwin MacBook 15.6.0 Darwin Kernel Version 15.6.0: Mon Jan  9 23:07:29 PST 2017; root:xnu-3248.60.11.2.1~1/RELEASE_X86_64 x86_64

bjorn3:
`x86_64-elf-gcc: command not found`

adrianN:
So it looks like the bootstrap script quit after trying to install git 

~~~~~
Installing missing packages...
git already exists at /usr/bin/git, no need to install git...
MacBook:~ aneumann$
~~~~~

and didn't install all the other packages

bjorn3:
brew exits with non zero error code when program is already installed. there is a redox issue for this.

jackpot51:
If this is fixed by your PR @adrianN, can you close?

PR #857, opened 2017-02-24
Ignore exit code from which in install_macos_pkg

adrianN:
On my OS X El Capitan machine `which` returns 1 if it can't find the binary. This kills the bootstrap script at the crucial moment when it detects that something needs to be installed. This change ignores unsuccessful exit codes from `which`. The man page for my `which` speaks of a `-s` switch that could be used instead of the `if [ -z`, but I don't know how portable this is.

This fixes Issue #850.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -26,7 +26,7 @@ install_macos_pkg()
         BIN_NAME=$PKG_NAME
     fi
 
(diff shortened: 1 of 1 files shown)


jackpot51:
Cool, looks reasonable.

Issue #852, opened 2017-02-23
bootstrap.sh help message wrong


ConnyOnny:
It says in the example usage "./bootstrap.sh -b buddy -e qemu", but I think there is no "-b" switch (any more?).

jackpot51:
This has been fixed in #863

PR #863, opened 2017-03-04
Remove erroneous -b option from bootstrap.sh help

SpamapS:
The help also tells you to use './bootstrap.sh' but the file wasn't +x in
the repo, so I've set that +x as well.

This closes issue #852


jackpot51:
Good change!

PR #887, opened 2017-03-23
Bootstrap script improvements for arch linux

neunenak:
I'm an arch linux user and I wasn't super-happy with how this script was
handling detecting missing packages and running updates that asked for
my sudo password without explaining what the script was doing. My
changes only run sudo pacman -Sy <package list> if necessary, and don't
do a full pacman -Syu since that's a pretty major system administration
operation.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -140,30 +140,38 @@ osx_homebrew()
 archLinux()
 {
 	echo "Detected Arch Linux"
-	echo "Updating system..."
-	sudo pacman -Syu
 
-	if [ -z "$(which nasm)" ]; then
-		echo "Installing nasm..."
-		sudo pacman -S nasm
+    local needed_packages=()
+
+    which nasm > /dev/null
+	if [ $? -ne 0 ]; then
+        $packages+="nasm"
 	fi
 
-	if [ -z "$(which git)" ]; then
-		echo "Installing git..."
-		sudo pacman -S git
+    which git > /dev/null
+	if [ $? -ne 0 ]; then
+        needed_packages+="git"
 	fi
+
+    which fusermount > /dev/null
+    if [ $? -ne 0 ]; then
+        needed_packages+="fuse2"
+    fi
 
 	if [ "$1" == "qemu" ]; then
-		if [ -z "$(which qemu-system-x86_64)" ]; then
-			echo "Installing QEMU..."
-			sudo pacman -S qemu
-		else
-			echo "QEMU already installed!"
+        which qemu-system-x86_64 > /dev/null
+		if [ $? -ne 0 ]; then
+            needed_packages+="qemu"
 		fi
 	fi
 
-	echo "Installing fuse..."
-	sudo pacman -S fuse
+    if [ ${#needed_packages[@]} -ne 0 ]; then
+        echo "Need to install the following packages: " $needed_packages
+        echo "Going to run 'sudo pacman -Sy $needed_packages'"
+        sudo pacman -Sy $needed_packages
+    else
+        echo "All needed system packages are installed"
+    fi
 }
 
 ###############################################################################


zarevucky:
As a fellow arch linux user, I feel obligated to point out that Arch Linux explicitly does not support partial upgrades, https://wiki.archlinux.org/index.php/System_maintenance#Partial_upgrades_are_unsupported. In other words, "pacman -Sy package" is an unsafe operation that can result in packages being broken. "pacman -Syu package" is the only reliable way to add a package to a system.

zarevucky:
As a further note, Arch Linux does not really support stable installations where you want not to upgrade stuff for some time. "pacman -Syu" is not a major system operation, it's a daily cron job.

polymetric1:
I've never really been a fan of this bootstrap script, to be honest. It just does stuff you're going to have to do (or learn how to do) eventually anyway. I think it's better practice to just get used to pulling, updating submodules, and installing/updating dependencies manually than to have a not-so-well written script do it.  
Furthermore, the guy who commited it seemed to be a bit inexperienced, and has basically not done anything on github since then. You can see the origins of this file in the [old repository](https://github.com/redox-os/old/blob/old_kernel/bootstrap.sh).

DomThePorcupine:
"the guy" here, you are right I was inexperienced. I got super excited about the project and after hanging around for a while and watching the plethora of issues open up in which people were unable to build because of dependencies/git issues, I figured I could maybe help that along. (Whether the issues stemmed from poor documentation at the time or something else I don't know) Furthermore there were less dependencies and less operating systems supported, when it was first written. Then things like fuse started being required and macOS support got a little wonky and that was made worse with the upgrade from Capitan to Sierra. Since it's origins I think it has become a little unwieldy, especially with so many people fiddling and adding specifics here and there. (I understand moreso now that many people have very specific packages installed/maintained on their machines, and you should be careful with package manager automation) One of the most recent changes to support macports has basically broken it's functionality for macOS. I think it was an awesome learning experience for me I'm sorry I could not contribute more as you seem to imply I should have, but perhaps you are right, it might be time to end bootstrap's life. Documentation has improved and the community is a little bigger now so there are more people to help out. Of course final say goes through @jackpot51

neunenak:
If the right thing to do here is actually kill that entire script and change the build preparation process, I'm happy to withdraw my pull request. I'm (very) new to redox development, that script was the first thing I ran when I tried to build it for the first time and I was just put off by the fact that it was asking for my sudo password without explaining what it was going to do with it.

As someone who (hopefully) will be building Redox a lot more in the near future, I'd like to ask what the initial build process from scratch *ought* to look like. I'd be happy to communicate with the relevant people in the project and put together a pull request that actually makes a good experience instead of just making minor improvements to this script.

bootstrap.sh
```diff
@@ -140,30 +140,38 @@ osx_homebrew()
...
-	sudo pacman -Syu
 
-	if [ -z "$(which nasm)" ]; then
-		echo "Installing nasm..."
-		sudo pacman -S nasm
+    local needed_packages=()
+
+    which nasm > /dev/null
+	if [ $? -ne 0 ]; then
+        $packages+="nasm"
```
sadika9: Did you mean `needed_packages+="nasm"`?

polymetric1:
@DomThePorcupine Crap, sorry, I might have been a bit harsh :P It's nothing personal. It's definitely better than any of the shell scripts I've made.

NickeZ:
I think we need to remove the package installations from the script. Here are two examples of pretty big rust projects and they both let the developer install the system dependencies manually: https://github.com/servo/servo https://github.com/jwilm/alacritty 

I think there is a lot of benefit to have a servo like workflow. But the whole `mach` script is probably overkill or impossible to maintain..

This is what you do to rebuild servo and I don't think it has to be harder than this:

```
git pull
./mach build [--release]
```
`./mach` needs to do a lot of things, like make sure you have the right rustc (it should be built from the same sources that are vendored into redox). From the top of my head, it needs to do `git pull --recursive --init` and it needs to do `cargo update`.

Coder206:
@NickeZ Yes, the command is `./mach build [-r|--release]`.

jackpot51:
Until you fix the nasm issue mentioned in a review, this cannot be merged

PR #896, opened 2017-03-27
Use absolute paths to virtualbox options

NickeZ:
According to the (virtualbox docs)[https://www.virtualbox.org/manual/ch08.html#vboxmanage-modifyvm] the file paths to `--uartmode` and `--nictracefile` must be absolute. `make virtualbox` fails for me without this patch.


jackpot51:
Seems reasonable.

NickeZ:
Oops, it wasn't my intention to actually change the names of the files.. I was in a bit of a hurry when I sent the PR..

PR #898, opened 2017-03-27
tcpd: fix dup with address

ids1024:


jackpot51:
Thanks! How did you find this?

ids1024:
A simple test program in C worked with opening the address directly, but not when opening `tcp:` and opening the address with the `dup` system call. So I checked what difference there was between the code in `tcpd`'s `open()` and `dup()`.

PR #908, opened 2017-04-06
Make easy targets section more readable.

PeterZhizhin:
For me (and as I think, for many others) it was very difficult to distinguish tasks that are proposed for Rust fluent developers (as well as OS and non-OS experts) and Rust newbies.

Add subsections for each type of developers.

- Added subsections in "Low-hanging fruit" section.


jackpot51:
Thanks!

PR #910, opened 2017-04-08
tcpd: Fix behavior when reading to small buffer

ids1024:


jackpot51:
I prefer iteration with the original method in scheme handlers - it usually generates better machine code

ids1024:
It's kind of awkward, since it needs to remove the initial elements of the `Vec`. I thought `.drain()` is probably the most efficient way to do that (though I haven't checked the assembly output).

It might be better to replace the `Vec` with a `VecDeque`, but it's in the `Tcp` struct in netutils, and changing it to a `VecDeque` breaks things.

jackpot51:
Fair enough. I see how drain is useful

PR #912, opened 2017-04-09
Build on NixOS

ksf:
**Problem**: Doesn't build on NixOS out of the box
**Solution**: Some nix expressions, a couple of changes to shebangs, setting some binary paths from outside the build system

**Blocking/related**: Needs https://github.com/redox-os/libc-artifacts/pull/6

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -64,3 +64,6 @@
 [submodule "programs/init"]
 	path = programs/init
 	url = https://github.com/redox-os/init.git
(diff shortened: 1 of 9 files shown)


ksf:
Huh. Apparently, I do *not* have the email address "foo@bar.baz" as my ~/.gitconfig claimed until seconds ago, in case anyone wonders about why the patches aren't attributed to me.

PR #961, opened 2017-06-10
Make build/filesystem.bin and build/initfs.tag targets .PHONY

ids1024:
This needs to be run to check for updates in the cookbook.

A more ideal solution would only rebuild these files if the cookbook updated a package; I don't know how best to implement that though.

### DIFF
--- a/mk/filesystem.mk
+++ b/mk/filesystem.mk
@@ -40,3 +40,5 @@ unmount: FORCE
 	sync
 	-$(FUMOUNT) build/filesystem/ || true
 	rm -rf build/filesystem/
+
+.PHONY: build/filesystem.bin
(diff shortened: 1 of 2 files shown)


jackpot51:
I would prefer a rebuild target that removes those two files, and then runs `make`

ids1024:
I don't know; it seems rather unintuitive if `make` ignores changes to userspace code.

jackpot51:
You could add FORCE to the prerequisites of build/initfs.tag and build/filesystem.bin

Phony is for targets that are not files

ids1024:
I know that's the primary use of `.PHONY`, but it seems appropriate here.

I haven't heard of `FORCE`, but the GNU make manual says this about it (https://www.gnu.org/software/make/manual/html_node/Force-Targets.html):

> Using ‘.PHONY’ is more explicit and more efficient. However, other versions of make do not support ‘.PHONY’; thus ‘FORCE’ appears in many makefiles.

Issue #963, opened 2017-06-11
OSX: undefined reference to `c_cpuid': make {virtualbox,qemu}


adarqui:
**Reproduction**: bootstrap.sh; make qemu (or make virtualbox)

**Expected behavior**: Don't error.

**Actual behavior**: I receive the following error:

```
redox$ make virtualbox
x86_64-elf-ld --gc-sections -z max-page-size=0x1000 -T kernel/linkers/x86_64.ld -o build/kernel build/libkernel.a
build/libkernel.a(raw_cpuid-9cf6b6e4636721a8.0.o): In function `raw_cpuid::CpuId::new::h7dfb0a7093d5537d':
raw_cpuid.cgu-0.rs:(.text._ZN9raw_cpuid5CpuId3new17h7dfb0a7093d5537dE+0x35): undefined reference to `c_cpuid'
raw_cpuid.cgu-0.rs:(.text._ZN9raw_cpuid5CpuId3new17h7dfb0a7093d5537dE+0x35): relocation truncated to fit: R_X86_64_PLT32 against undefined symbol `c_cpuid'
build/libkernel.a(raw_cpuid-9cf6b6e4636721a8.0.o): In function `raw_cpuid::CpuId::get_vendor_info::hd4ca29f2f98219c9':
raw_cpuid.cgu-0.rs:(.text._ZN9raw_cpuid5CpuId15get_vendor_info17hd4ca29f2f98219c9E+0x39): undefined reference to `c_cpuid'
raw_cpuid.cgu-0.rs:(.text._ZN9raw_cpuid5CpuId15get_vendor_info17hd4ca29f2f98219c9E+0x39): relocation truncated to fit: R_X86_64_PLT32 against undefined symbol `c_cpuid'
build/libkernel.a(raw_cpuid-9cf6b6e4636721a8.0.o): In function `raw_cpuid::CpuId::get_feature_info::h7b77103b518d3f0e':
raw_cpuid.cgu-0.rs:(.text._ZN9raw_cpuid5CpuId16get_feature_info17h7b77103b518d3f0eE+0x3e): undefined reference to `c_cpuid'
raw_cpuid.cgu-0.rs:(.text._ZN9raw_cpuid5CpuId16get_feature_info17h7b77103b518d3f0eE+0x3e): relocation truncated to fit: R_X86_64_PLT32 against undefined symbol `c_cpuid'
[...]

PR #995, opened 2017-07-13
Tweaks to Makefiles to be able to build the kernel on macOS again.

goyox86:
**Problem**: 

The kernel can't be built (without fiddling with env vars and stuff which) due to one of `raw-cpuid` C dependencies being compiled as a Mach-O object file by the macOS toolchain and also the cookbook scripts assume GNU utilities. This may be a show stopper for newcomers on macOS.

**Solution**: 

- Force using the macOS Redox toolchain compiler and archiver in order
to generate a proper ELF file when building the 'raw-cpuid'
crate C code.

- Add the `findutils` and `coreutils` Homebrew/MacPorts
packages to the bootstrap script in order to setup the proper
environment for the cookbook scripts (as they assume GNU tools).
This should be temporary until the cookbook scripts get migrated
to Rust. `findutils` contains `gfind` AKA GNU find. And `coreutils` contain
`stat` which may look overkill but I guess after this we can safely assume
that cookbook scripts will have GNU coreutils available making them simpler.

- Force CC and AR in kernel buidling/linking on macOS.
- Add the `findutils` and `coreutils` Homebrew/MacPorts
packages to the bootstrap script.

**Fixes**: https://github.com/redox-os/redox/issues/963

**State**: [the state of this PR, e.g. WIP, ready, etc.]

**Blocking/related**: I'm gonna link the cookbook PR soon.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -93,6 +93,8 @@ osx_macports()
         install_macports_pkg "virtualbox"
     fi
 
+    install_macports_pkg "coreutils"
+    install_macports_pkg "findutils"
     install_macports_pkg "gcc49" "gcc-4.9"
     install_macports_pkg "nasm"
     install_macports_pkg "pkgconfig"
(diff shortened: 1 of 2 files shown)


jackpot51:
Looks good to me

PR #1004, opened 2017-07-23
Add a mention for selinux systems in Docker README

fengalin:
**Problem**: 
Selinux blocks the following:
- gosu execution labeling (no selinux policy).
- permission is denied to the container's user on the tree source binding
under /src. Using a different location doesn't solve the issue.

**Solution**:
The workaround consists in disabling selinux labeling when running the container.


jackpot51:
Thank you!

PR #1005, opened 2017-07-24
Build docker container with ready to use user env

fengalin:
**Problem:** In order to initialize the user with an id matching the host's user id, a `-e LOCAL_USER_ID` argument is passed to the run command. The `entrypoint.sh` script is then in charge of creating the user and changing ownership on `cargo` and `rustup` directories. Finally, `gosu` runs the requested command (`make all`) as `user`.
This strategy allows adapting the container to the environment of the running user. This is needed as we bind `redox` source tree to the container and need read/write access to it. However, every time the container is run, the user creation and `chown -R` are executed.

**Solution:** Pass `--build-arg` to the container build command in order to create the user at container build time. Then, everything can be installed like a local host user would. In order to preserve the ability to adapt the container to the environment of the running user, the host's user id must be checked at run time and the container user id changed when necessary. Take this opportunity to adapt the group id.

**Changes introduced by this pull request:**
- Move user environment creation from `entrypoint.sh` to `Dockerfile`.
- Change user id and group id if necessary.
- `apt-get update` all the repos only once. Update `redox` repo individually.
- Add `user` to the list of `sudoers`.
- Use default user directories for cargo and rust.
- Bind the source tree to the user's `~/src`. Run command uses `/home/user/src/` instead of `/src`. See `README.md`.
- Update `README.md`: update submodule sync step to avoid a dependency on cargo.
- Update `README.md`: use `make update all` to update cargo's environment in the container.

**Constraint**: `x86-64-elf-redox-*` packages use `apt-transport-https`, so `redox.list` must be added after the first `apt-get install`.

**State**: Tested on Fedora 26 only.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -4,11 +4,13 @@
 
 ```shell
 git clone https://github.com/redox-os/redox.git ; cd redox #1
-make pull #2
-docker build -t redox docker/ #3
+git pull --rebase --recurse-submodules && git submodule sync \
+    && git submodule update --recursive --init #2
+docker build --build-arg LOCAL_UID="$(id -u)" --build-arg LOCAL_GID="$(id -g)" \
+    -t redox docker/ #3
 docker run --cap-add MKNOD --cap-add SYS_ADMIN \
-    --device /dev/fuse -e LOCAL_USER_ID="$(id -u)" \
-    -v "$(pwd):/src" --rm redox make all #4
+    -e LOCAL_UID="$(id -u)" -e LOCAL_GID="$(id -g)" \
+    --device /dev/fuse -v "$(pwd):/home/user/src" --rm redox make update all #4
 make qemu #5
 ```
 To unpack:
@@ -21,7 +23,7 @@ To unpack:
 On selinux systems, replace #4 with:
 ```
 docker run --cap-add MKNOD --cap-add SYS_ADMIN \
-    --device /dev/fuse -e LOCAL_USER_ID="$(id -u)" \
-    -v "$(pwd):/src" --security-opt label=disable \
-    --rm redox make all
+    -e LOCAL_UID="$(id -u)" -e LOCAL_GID="$(id -g)" \
+    --device /dev/fuse -v "$(pwd):/home/user/src" --security-opt label=disable \
+    --rm redox make update all
 ```
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,35 +1,44 @@
 FROM ubuntu:17.04
 
 ENV REDOX_TOOLCHAIN_APT http://static.redox-os.org/toolchain/apt/
-ENV SRC_PATH /src
-ENV CARGO_HOME /cargo
-ENV RUSTUP_HOME /rustup
-ENV PATH $CARGO_HOME/bin:$PATH
 
-RUN apt-get update \
+ENV USER user
+ARG LOCAL_UID=local
+ARG LOCAL_GID=local
+ENV BUILD_UID=${LOCAL_UID:-9001}
+ENV BUILD_GID=${LOCAL_GID:-9001}
+
+RUN   apt-get update \
       && apt-get install -y git gosu gcc fuse nasm qemu-utils pkg-config \
-         libfuse-dev make curl file sudo apt-transport-https autoconf flex \
-         bison texinfo \
-      && mkdir -p $CARGO_HOME \
-      && mkdir -p $RUSTUP_HOME \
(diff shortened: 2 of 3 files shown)


batonius:
I like the idea, I've too noticed that recursive chown is slow, but we should keep in mind that this approach would result in problems if an image built by a user was to be used by another user, which is highly unlikely but possible. For the same reason, it will be hard to use DockerHub if we decide to distribute the image in the future since it pre-builds images IIRC.

Also, I believe you can just get rid of `entrypoint.sh` since it does nothing now.

fengalin:
Good point @batonius. I didn't figure that out.
I can get rid of `entrypoint.sh` if this PR is still going to be merged.

fengalin:
It seems like we can get the best of both worlds. From [this page](https://docs.docker.com/engine/reference/run/#user), I found:
> The developer can set a default user to run the first process with the Dockerfile USER instruction. When starting a container, the operator can override the USER instruction by passing the -u option.

I'll take a look at removing `entrypoint.h` tomorrow. I'm new to docker, so if anyone knows better, please let me know.

batonius:
The problem here is that the user should have the same UID as the current host user for file ownership to be correct, so you can't just use the pre-existing user in every case.

An alternative would be to keep both old and new mechanisms and change `entrypoint.sh` in such a way that it does nothing if LOCAL_USER_ID is the same as UID of the user that was created during the build, but creates a new user and changes the ownership overwise.

fengalin:
Yeah, I guess I hoped for some magic with docker's `-u` option, but of course that's not enough.
I think I got it right this time. See last commit.

docker/README.md
```diff
@@ -1,14 +1,15 @@
 ### Building Redox using Docker images with the toolchain
 
-*All you need is git, make, qemu, fuse and docker. The method requires a non-privileged user able to run the `docker` command, which is usually achieved by adding the user to the `docker` group.*
+*All you need is rustup, git, make, qemu, fuse and docker. The method requires a non-privileged user able to run the `docker` command, which is usually achieved by adding the user to the `docker` group.*
```
batonius: Why do we need rustup here? This is requirements for the host system, and we only use rust inside the image.

fengalin: I found that `make pull` uses `cargo`. I ran into the issue testing my container with a brand new user.

batonius: Oh, I see, I totally missed it. Maybe we should replace `make pull` with 
```bash
git pull --rebase --recurse-submodules
git submodule sync
git submodule update --recursive --init 
``` 
in the README, since all we need here is to pull all the submodules? There's no actual need for rust to be installed on the host system.

fengalin: Yes indeed. I'll do that.

fengalin: Done

fengalin: I changed my mind. What do you think about doing `make pull && make all` when running the container? This way, we get the `cargo` stuff from `make pull` within the container.

batonius: You're right, we don't need the submodules to build and run the image, so we can just do `make pull all` instead of `make all` when we run the container.
The only problem is that `make pull` calls `make clean`, so every build will be a complete rebuild. I think the README should mention that `pull` could be omitted for subsequent builds.

fengalin: Sorry, I removed my previous comment because I was facing an issue testing this and I didn't want to clutter the thread.

The issue was the following, when I ran the container with `make pull && make all` or just `make update && make all`, I ended up with `xargo` not being found any more. It seems like your suggestion to `make pull all` instead does the trick.

However, syncing the submodules from the container doesn't seem to be a good idea. I found the sync was very long. In the network monitor, I noticed as much data being uploaded as dowloaded. My guess is that the communication between the host and the container uses the network stack. So I guess we are better off syncing from the host, as you proposed before.

So my last proposal is to use the new step 2 you proposed (already commited) and change the run command to `make update all`. This way, we won't execute `make clean` every time and will keep the `README` simple.

fengalin: Done.
Thank you for your support!

batonius: I'm not sure why syncing submodules from within a container would be significantly slower than doing it from a host system. I've tested it and it took ~10 minutes to do `make pull` on a fresh repo from within a container, approximately the same time it takes on the host system. Unfortunately,  I've run into some github authentification problems while doing this because I was using my own fork. So, in the end, it looks like syncing submodules from the host system is indeed the best way for now.

> I ended up with xargo not being found any more. 

I guess you didn't use quotes, so your shell thought `make all` was a separate command and tried to run it on your host machine directly.

PR #1030, opened 2017-08-16
Docker: align APT key retrieval with Cookbook README

fengalin:
Use the same command as the Cookbook README for Redox OS APT key retrieval.

Keep adding the Redox OS repository in a dedicated file. This allows updating it specifically without updating the other repositories one more time.


jackpot51:
Thanks!

PR #1044, opened 2017-08-29
Docker: Cache cargo downloads with named volume

andre-richter:
**Problem**:
Currently, cargo downloads dependencies on every single run.

**Solution**:
Get rid of this overhead by caching /home/user/.cargo in a named volume.

**Changes introduced by this pull request**:
Changes in docker/README.md

**Drawbacks**:
Some additional disk space for the named volume is occupied.

**TODOs**:

**Fixes**:

**State**: 
ready

**Blocking/related**:

**Other**:

PR #1043, opened 2017-08-29
Docker: allow building the container on MacOS and Linux

fengalin:
On MacOS, while building the docker container, `useradd` returns with code 4 (UID already in use), even with option -o. On this platform, the access rights for a volume attached to a running container are translated to the container's user and group. So, there is no need to handle UID and GID compliance like we do on Linux. See this for reference: https://docs.docker.com/docker-for-mac/osxfs/

This modification allows not specifying the UID and GID while building the container, keeping the defaults for `useradd`.

Fixes #1037

PR #1045, opened 2017-09-02
Rework Docker

andre-richter:
Hi all,

I took the liberty to rework some of the docker stuff. Here's a list of changes and why I think they are useful.

**1. Use the official Rust nightly docker image as base**
Recently, an official docker image for Rust has emerged. Switch to it to be more aligned with upstream.

**2. Remove hardcoded user**
Run-time information of any kind should not be stored in prebuilt container images. This makes it easier to distribute images to all different kinds of client OSes e.g. via dockerhub (Let's start work on that soon). I saw that this has been done on purpose in #1005 to avoid time-consuming `chown -R` runs on every container startup.
In this rework, named volumes are used to cache the rust toolchain (see next point). This, in turn, enables caching of the `chown` run, which allows to again shift user creation to runtime.
 
**3. Use named volumes to cache `.cargo` and `.rustup` toolchain folders**
Has been already introduced in #1044 for the cargo folder. Extended to the rustup folder as well. Provides additional benefits as described in point 2.

**4. Unify workflows for Linux and MacOS**
Both Linux and MacOS can use the same commands now to build and run the containers, including passing of `LOCAL_UID` and `LOCAL_GID` env variables. 
I know only recently this has been separated in #1043 due to an error of `groupadd`. Since we shifted user creation back to runtime, container building should not be affected anymore.
Additionally, I tested this patch on MacOS, and `groupadd --non-unique --gid $RUN_GID $USER_NAME` does not fail if `RUN_GID` equals `20`, like it is the case on MacOS. The duplicate group is created without errors.
@adrianbrink, maybe you can have a try too?

**5. Cleanup apt folders after installing**
Shrinks the container a bit

**6. Make bash the default fallback command**
Simplifies launching an interactive session via `docker run -it`

**7. Add guide for launching an interactive session**
Also make the prompt of the interactive session pretty and indicate to the user that he is in a containerized shell.
This is intended for workflows where people leave the interactive shell open in order to have fast access to the container toolchain.

**State:** Tested on Ubuntu 16.04 and MacOS Sierra.


jackpot51:
Looks really good to me!

PR #1046, opened 2017-09-03
docker: Adapt README for dockerhub usage

andre-richter:
- Add `redoxos/` namespace
- Move container building to end of README as fallback.


docker/README.md
```diff
@@ -34,7 +32,7 @@ docker run --cap-add MKNOD --cap-add SYS_ADMIN --device /dev/fuse \
     -e LOCAL_UID="$(id -u)" -e LOCAL_GID="$(id -g)" \
     -v redox-"$(id -u)-$(id -g)"-cargo:/usr/local/cargo \
     -v redox-"$(id -u)-$(id -g)"-rustup:/usr/local/rustup \
-    -v "$(pwd):$(pwd)" -w "$(pwd)" --rm redox make fetch all
+    -v "$(pwd):$(pwd)" -w "$(pwd)" --rm redoxos/redox make fetch all
```
fengalin: If this is to be pushed to dockerhub, I would use another name than `redox` for this image as it's an image of the cross compilation toolchain, not the OS.

fengalin:
These PR are great steps ahead!
I made a few comments.

docker/README.md
```diff
@@ -69,3 +67,10 @@ docker run --cap-add MKNOD --cap-add SYS_ADMIN --device /dev/fuse \
 docker volume rm redox-"$(id -u)-$(id -g)"-cargo \
```
fengalin: I think this section should be extended to the rustup volume.

docker/README.md
```diff
@@ -34,7 +32,7 @@ docker run --cap-add MKNOD --cap-add SYS_ADMIN --device /dev/fuse \
     -e LOCAL_UID="$(id -u)" -e LOCAL_GID="$(id -g)" \
     -v redox-"$(id -u)-$(id -g)"-cargo:/usr/local/cargo \
     -v redox-"$(id -u)-$(id -g)"-rustup:/usr/local/rustup \
-    -v "$(pwd):$(pwd)" -w "$(pwd)" --rm redox make fetch all
+    -v "$(pwd):$(pwd)" -w "$(pwd)" --rm redoxos/redox make fetch all
 ```
 #### Linux with security modules<br>
```
fengalin: Would you mind removing the `<br>` I left behind? :)

docker/README.md
andre-richter: Hi, thanks for the feedback.

Regarding this comment, @jackpot51 already set it up that way: https://hub.docker.com/r/redoxos/

In my opinion, the naming is okay. Redox is an OS with a non-linux kernel, so the userspace parts could not boot within within a docker container anyways. Hence not so much potential for confusion here.

How would you name it? `redox-toolchain`?

docker/README.md
andre-richter: Not sure I get what you mean. Can you elaborate?

docker/README.md
andre-richter: Force pushed a new version without it.

docker/README.md
fengalin: Yes, you're right. I guess `redox-toolchain` would be redundant considering the namespace, maybe something like `cross-toolchain`.

docker/README.md
fengalin: People might also want to remove the volume `redox-"$(id -u)-$(id -g)"-rustup`.

I guess the `Clear the named volume containing the cargo cache` section could be generalized to `Clear the named volumes`. And a command added to clean the volume `redox-"$(id -u)-$(id -g)"-rustup`.

docker/README.md
fengalin: Thanks

docker/README.md
andre-richter: Ah, it was already there in the next line, the preview window here cut it out. The headline was referring to cargo only though.
Fixed it.

fengalin:
LGTM.

@jackpot51 what do you think about this: https://github.com/redox-os/redox/pull/1046#pullrequestreview-60299102?

docker/README.md
andre-richter: This is nitpicking, but is it a cross toolchain after all? The docker container is x86_64 and that is also what the target is compiled for.

`redoxos/toolchain` ?

fengalin: I think we can talk about a cross toolchain as it uses a Linux host to build Redox binaries.

But as you pointed out earlier, the fact that it is a docker image makes it obvious that the host is not Redox. Under this perspective, `toolchain` is probably enough for people to understand.

jackpot51:
Looks good! We can rename it in the future

Issue #1052, opened 2017-09-06
OpenSUSE detected as Ubuntu/Debian


krrg:
**Reproduction**:
Attempt to run the bootstrap.sh script on an OpenSUSE 42.3 installation.

**Expected behavior**:
Should detect as a SUSE-based installation and run the suse() function.

**Actual behavior**:
Instead, detects as a Debian/Ubuntu system and uses `apt-get`.  Some package names are incorrect, however.

The `zypper-aptitude` plugin is evidently installed by default on OpenSUSE 42.2.  This plugin provides an `apt-get` command that functions very similarly to the Debian `apt-get`.  This causes the bootstrapping script to mistake a SUSE installation for a Debian installation.  The installation will fail because OpenSUSE is unable to find some of the specific packages called for (such as `build-essential`).

A potential fix for this would be to change the order of the checks; check if `zypper` is present first, and only if it is not, check for `apt-get`.  PR forthcoming.

PR #1053, opened 2017-09-06
Update bootstrap.sh

krrg:
See https://github.com/redox-os/redox/issues/1052

This is just something fast I threw together to get the general idea across of what might fix the problem.  Basic idea is to test SUSE before testing for Ubuntu.

### DIFF
--- a/bootstrap.sh
+++ b/bootstrap.sh
@@ -512,26 +512,29 @@ else
 	if hash 2>/dev/null pacman; then
 		archLinux "$emulator"
 	fi
-	# Debian or any derivative of it
-	if hash 2>/dev/null apt-get; then
-		ubuntu "$emulator" "$defpackman"
-	fi
-	# Fedora
-	if hash 2>/dev/null dnf; then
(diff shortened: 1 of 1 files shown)


bootstrap.sh
```diff
@@ -512,26 +512,29 @@ else
...
-	# Debian or any derivative of it
-	if hash 2>/dev/null apt-get; then
-		ubuntu "$emulator" "$defpackman"
-	fi
-	# Fedora
-	if hash 2>/dev/null dnf; then
-		fedora "$emulator"
-	fi
 	# Suse and derivatives
 	if hash 2>/dev/null zypper; then
```
vbuell: Wouldn't it be more correct to convert all ifs into elifs? The fact that it's written in such way that few may be invoked at the same time smells as for me.

krrg: Yes, that's an excellent point, let me update the PR.

jackpot51:
Looks good to me!

Issue #1102, opened 2017-11-13
/usr inappropriately soft-links to /: Results in infinite recursion with, "ls -lR /"


gbutler69:
- [ X] I agree that I have searched opened and closed issues to prevent duplicates.

--------------------

## Description

If you go to the root of the file-system in Redox and enter the command, `ls -lR` (or simply type, `ls -lR /` anywhere) it goes into infinite recursion showing files/folders in:

/usr/bin
/usr/usr/bin
/usr/usr/usr/bin
...
/usr/usr/usr/.........../usr/bin
etc.

## Environment info

Inside Redox OS running under QEMU following the proscribed procedures.

- Operating system:

Ubuntu 17.04 (Linux)

- `uname -a`:

`Linux butler-Leopard-Extreme 4.10.0-38-generic #42-Ubuntu SMP Tue Oct 10 13:24:27 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux`

- `rustc -V`:

`rustc 1.23.0-nightly (79cfce3d3 2017-11-12)`

- `git rev-parse HEAD`:

`5f7233417b7934604ac5e355989dd4613c1b64c5`

## Steps to reproduce

   1. Build Redox
   2. start it with, "make qemu"
   3. Login using "root/password"
   4. enter the following command, "ls -lR /"
   5. watch as it infinitely recurses into /usr, /usr/usr, /usr/usr/usr, ...

## Behavior

- **Expected behavior**:

   * lists all files/folders on the file-system without infinitely recursing

- **Actual behavior**:

    * infinite recursion into /usr, /usr/usr/, ...

- **Proposed solution**:

    Change /usr to be an actual folder and /usr/bin to be a soft-links to /bin  as follows:

````
rm /usr
mkdir /usr
chown root:root /usr
chmod 755 /usr
ln -s /bin /usr/bin
````

EDIT: Make the sub-folders of /usr that are soft-links correctly reflect the desired/current folders at the / level in RedoxOS

xTibor:
Here's where the recursive symlink is defined:
https://github.com/redox-os/redox/blob/f9865f49d1e597f54beb61dbe456a0969396dc7a/filesystem.toml#L177-L180

gbutler69:
I'm working on this. Already have a fix ready. Just working on submitting PR.

PR #1103, opened 2017-11-13
Corrected filesystem.toml to correctly soft-link folders in /usr to f…

gbutler69:
…olders in / so as to avoid infinite recursiong with 'ls -lR /' as per https://github.com/redox-os/redox/issues/1102

- [...]
- [...]
- [...]

**Drawbacks**: [if any, describe the drawbacks of this pull request.]

**TODOs**: [what is not done yet.]

**Fixes**: [what issues this fixes.]

**State**: [the state of this PR, e.g. WIP, ready, etc.]

**Blocking/related**: [issues or PRs blocking or being related to this issue.]

**Other**: [optional: for other relevant information that should be known or cannot be described in the other fields.]

------

_The above template is not necessary for smaller PRs._

### DIFF
--- a/filesystem.toml
+++ b/filesystem.toml
@@ -175,6 +175,32 @@ Welcome to Redox OS!
 """
 
 [[files]]
-path = "/usr"
-data = "/"
+path = "/usr/bin"
+data = "/bin"
 symlink = true
+
+[[files]]
+path = "/usr/sbin"
+data = "/sbin"
+symlink = true
+
+[[files]]
+path = "/usr/var"
+data = "/var"
+symlink = true
+
+[[files]]
+path = "/usr/share"
+data = "/share"
+symlink = true
+
+[[files]]
+path = "/usr/include"
+data = "/include"
+symlink = true
+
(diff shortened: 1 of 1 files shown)


filesystem.toml
```diff
@@ -175,6 +175,32 @@ Welcome to Redox OS!
...
+data = "/bin"
 symlink = true
+
+[[files]]
+path = "/usr/sbin"
+data = "/sbin"
+symlink = true
+
+[[files]]
+path = "/usr/var"
```
ids1024: Do some systems have a `/usr/var`? Archlinux doesn't seem to have such a directory.

gbutler69: No, I don't think they do come to think of it. For some reason I thought Linux did, but, just verified it doesn't. I'll remove that. Please reject this PR and let me re-work it and I'll submit a new PR Shortly.

gbutler69:
I'm closing this PR because I need to re-work the solution. The initial solution I proposed isn't going to work.

PR #1104, opened 2017-11-14
Fix issue #1102: Infinite Recursion with "ls -lR /" in RedoxOS

gbutler69:
**Problem**: see issue #1102 

**Solution**: see solution as described in issue #1102 

- /usr is no longer a soft-link to /, instead, /usr/bin links to /bin etc (same for all folders/files in / corresponding soft-links are created in /usr that link to them) - see #1102 

**Drawbacks**: Any new top-level files/folders added in the future may need corresponding soft-links created in /usr if /usr should always exactly be a mirror of the top-level folder (questionable???)

**TODOs**: nothing that I'm aware of

**Fixes**: see issue #1102 

**State**: READY

**Blocking/related**:  none that I'm aware of

**Other**: (n/a)

### DIFF
--- a/filesystem.toml
+++ b/filesystem.toml
@@ -175,32 +175,6 @@ Welcome to Redox OS!
 """
 
 [[files]]
-path = "/usr/bin"
-data = "/bin"
+path = "/usr"
+data = "/"
 symlink = true
-
-[[files]]
-path = "/usr/sbin"
-data = "/sbin"
-symlink = true
-
-[[files]]
-path = "/usr/var"
-data = "/var"
-symlink = true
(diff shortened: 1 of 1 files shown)


gbutler69:
OK, I see that the Travis build failed, but, I don't think the failure is related to this change/PR. Thoughts?

jackpot51:
You only need to link /usr/bin

gbutler69:
Closing this PR as @jackpot51 has indicated that it is not desired that all top-level folders be soft-linked from /usr, but, only /usr/bin -> /bin.

PR #1105, opened 2017-11-14
Fix issue #1102: Infinite Recursion with "ls -lR /" in RedoxOS

gbutler69:
Problem: see issue #1102

Solution: see solution as described in issue #1102

Changes introduced by this pull request:

    /usr is no longer a soft-link to /, instead, /usr/bin links to /bin  - see #1102

Drawbacks: Any new top-level files/folders added in the future may need corresponding soft-links created in /usr if it is intended that they be available through /usr (questionable???)

TODOs: nothing that I'm aware of

Fixes: see issue #1102

State: READY

Blocking/related: none that I'm aware of

Other: (n/a)

### DIFF
--- a/filesystem.toml
+++ b/filesystem.toml
@@ -175,32 +175,6 @@ Welcome to Redox OS!
 """
 
 [[files]]
-path = "/usr/bin"
-data = "/bin"
+path = "/usr"
+data = "/"
 symlink = true
-
-[[files]]
-path = "/usr/sbin"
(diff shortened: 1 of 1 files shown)


gbutler69:
OK, I believe this is correct now. Only /usr/bin -> /bin exists as a soft-link. No others. I don't believe the Travis build failure is related to this change.

jackpot51:
Cool!

PR #1114, opened 2017-12-02
Keep debug info in a .sym file

dlrobertson:
**Problem**:

When debugging the redox kernel, it is compiled without debug symbols
which makes debugging difficult.

**Solution**:

Compile with debug info and place debug info in separate `.sym` file.

 - Compile time changes
   - Compile libkernel.a with debug info
   - Copy debug info to kernel.sym
   - Strip the kernel of debug symbols
 - Workflow changes
   - Add `debug` option to `make qemu`

**Drawbacks**:

The `kernel.sym` can be large and `libkernel.a` is larger.
`kernel.sym` is typically around 6 MB for me. It might be
a good idea to put this behind a build option.

**Blocking/related**:

 - redox-os/kernel#64

**TODO**:

 - Write up a wiki page on debugging the redox kernel with `qemu`
   and `gdb`

**Try it out**:

To test this out first make sure that the redox-os/kernel#64 kernel is used,
then rebuild the kernel. After rebuilding the kernel, run the VM and use
`make qemu debug=yes`. Now it's time to crack open `gdb`

```
$ rust-gdb ./build/kernel
(gdb) target remote localhost:1234
(gdb) symbol-file ./build/kernel.sym
... other gdb commands
(gdb) continue
```

Go back to your VM and trigger the syscall you're interested in.

### DIFF
--- a/mk/kernel.mk
+++ b/mk/kernel.mk
@@ -1,9 +1,9 @@
 build/libkernel.a: kernel/Cargo.toml kernel/src/* kernel/src/*/* kernel/src/*/*/* build/initfs.tag
 # Temporary fix for https://github.com/redox-os/redox/issues/963 allowing to build on macOS
 ifeq ($(UNAME),Darwin)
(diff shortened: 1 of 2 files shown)


jackpot51:
Can you update the kernel submodule?

dlrobertson:
Np! updated

jackpot51:
Thanks!

PR #1123, opened 2017-12-18
Move null: and zero: to user space

Arcterus:
This should not be merged at the moment.  After redox-os/kernel#67 is merged, I will need to update the `kernel` and `cookbook` submodules so `nulld` and `zerod` can be used.

### DIFF
--- a/filesystem.toml
+++ b/filesystem.toml
@@ -39,6 +39,7 @@ netdb = {}
 netstack = {}
 netutils = {}
 #newlib = {}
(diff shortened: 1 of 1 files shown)


jackpot51:
Alright, we are ready for this!

Good job finding all the places that need to change and PRing them!

Arcterus:
I think everything's good now :)

jackpot51:
Cool - merged!

Issue #1124, opened 2017-12-19
ion crashes with a page fault after login


nihathrael:
- [x ] I agree that I have searched opened and closed issues to prevent duplicates.

--------------------

## Description

ion crashes with a page fault on login in the terminal or when starting obital in the UI.

## Environment info

- Arch Linux
- qemu and virtualbox

- Operating system:
Arch Linux
- `uname -a`:
`Linux reaper 4.14.6-1-ARCH #1 SMP PREEMPT Thu Dec 14 21:26:16 UTC 2017 x86_64 GNU/Linux`
- `rustc -V`:
`rustc 1.24.0-nightly (dc39c3169 2017-12-17)`
- `git rev-parse HEAD`:
`ee7b0462a37897545114f1bdd41dc48530ffc2f1` (HEAD at December 19th 2017)

## Steps to reproduce

1. Launch redox `make qemu vga=none`
2. login
3. see pagefault on launch of ion

OR

1. Launch redox `make qemu`
2. login to UI
3. launch orbital 
4. See page fault on terminal

## Behavior

- **Expected behavior**:
I can use Ion

- **Actual behavior**:
Pagefault

```
Page fault: 0000000000000000
RFLAG: 0000000000010246
CS:    0000000000000023
RIP:   000000000056E16F
CODE:  0000000000000006
RAX:   0000028000001000
RCX:   0000028000001000
RDX:   0000000000000A30
RDI:   00000080000005C0
RSI:   00000080000005F0
R8:    8080808080808080
R9:    0000000200000000
R10:   0000000000000000
R11:   0000000000000065
RBX:   0000000000000000
RBP:   00000180000FFC40
R12:   0000000000000019
R13:   000000000000002E
R14:   00000080000005C0
R15:   00000000005F6BE8
FS:    0000000000000033
TRACE: FFFFFF40002017B0
  FFFFFF40002017B0: FFFFFF000019C6B5
    FFFFFF000019C610+00A5
    kernel::arch::x86_64::interrupt::exception::page::inner::h55eaa041bbf4716d
  FFFFFF4000201810: FFFFFF000019C60B
    FFFFFF000019C5E0+002B
    kernel::arch::x86_64::interrupt::exception::page::h7b475d05caa98e7d
  00000180000FFC40: 0000000000575E86
  00000180000FFD30: 0000000000575B52
  00000180000FFD70: 000000000056E30C
  00000180000FFDC0: 00000000005763F9
  00000180000FFE80: 00000000004001F5
[...]

PR #1125, opened 2017-12-24
Use bash as shell until #1124 is fixed

NilSet:
Ion is currently broken on redox due to #1124 . This switches to bash for now.

### DIFF
--- a/filesystem.toml
+++ b/filesystem.toml
@@ -10,7 +10,7 @@ prompt = false
 #acid = {}
 #autoconf = {}
 #automake = {}
-#bash = {}
+bash = {}
 #binutils = {}
 #ca-certificates = {}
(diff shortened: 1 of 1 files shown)


ids1024:
I don't know what the cause of this issue is, but if it's due to a change in Ion, it would probably be better to temporarily pin the Ion recipe to an old working commit.

NilSet:
> Reverting ION as far back as august 6th doesn't help, still crashes
> Reverting Kernel back to 1.3.2 doesn't help, still crashes
I don't think it's going to be that simple. Probably some unpinned deep dependency.

jackpot51:
I agree that we should use bash for now, because in principle, we need an image that works all the time.

PR #1126, opened 2017-12-27
Create rust-toolchain

jD91mZM2:
Ring isn't compatible with latest nightly.
This add a rust-toolchain file to automatically downgrade nightly.
yay

Warning: Hasn't been tested because I don't have redox locally, but should in theory work


ids1024:
This seems to work, though you have to run `rustup override unset` for it to take effect, if you already have an override set.

PR #1129, opened 2017-12-30
Do not mention 'rustup override set' in README

ids1024:
The rust-toolchain file replaces this, and is probably better.


jackpot51:
Thanks!

PR #1130, opened 2018-01-06
Fix broken link to libstd

Nopey:
**Problem**: Libstd link 404s

**Solution**: Libstd link now points to github/Yoric/libstd instead of github/redox-os/libstd

**Alternatively**: libstd could be added to the organization


jackpot51:
libstd is coming directly from Rust now. This line should be removed instead of changed

Nopey:
Oh neat, I'll change my PR

Nopey:
Perhaps it should be labelled as abandoned and put with the old kernel at the bottom of the table?

jackpot51:
Yes, I think that would work

jackpot51:
Thanks

PR #1132, opened 2018-01-09
Docker: add autopoint

fengalin:
Needed by `xz`

This should solve https://github.com/redox-os/redox/issues/1131.


jackpot51:
Thanks

Issue #1176, opened 2018-04-02
'make all' failure requires PATH be set.


ianmiell:
FWIW, my install failed with this:

```
imiell@welles:~/redox/redox  ⑂ master  $  make all
mkdir -p build
nasm -f bin -o build/bootloader -D ARCH_x86_64 -ibootloader/x86_64/ bootloader/x86_64/disk.asm
cd kernel && xargo clean
/bin/sh: 1: xargo: not found
mk/initfs.mk:2: recipe for target 'build/initfs.tag' failed
make: *** [build/initfs.tag] Error 127
```

so I set:

```
PATH=${PATH}:/home/imiell/.cargo/bin
```

and make all worked.

PR #1177, opened 2018-04-02
Update build steps

xTibor:
After installing rustup the PATH variable needs to be updated for the current shell to be able to use cargo. The installer even warns about this. Possible fix for #1176.


rpm0372:
It would be nice if bootstrap.sh would source cargo env when bootstrap.sh installs rust so the user doesn't have to before proceeding to compile in the same terminal window.

jackpot51:
The bootstrap script could not do this unless it launches a shell itself

PR #1194, opened 2019-03-31
Add missing dependencies to fedora in bootstrap.sh

faern:
I had to install these extra packages on a clean Fedora 29 Workstation before `make all` could finish successfully.


longwusha:
@faern I suggest you go to the gitlab repository to mention this PR. Now this repo is only used as a mirror. No one will view and merge the code here.

jackpot51:
Yep, please submit to https://gitlab.redox-os.org/redox-os/redox

PR #1196, opened 2022-03-15
Very important commit

Iceberg-Dev:
Please be kind to mac users

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /build/
 /prefix/
 .config
+.DS_Store


bjorn3:
You can put it in your global gitignore instead. This applies for all repos. You can find it at `.config/git/ignore` or `Library/git/ignore` I think. By the way this is just a mirror of https://gitlab.redox-os.org/redox-os/redox

Iceberg-Dev:
Thank you

